PR #53, opened 2014-05-20
Initial read-only support for notes, style objects

ludoo:
I added initial support for footnotes/endnotes parts, and footnotes/endnotes references. The new classes are read-only, much in the same way the styles part is, as I only need to read existing documents.

I also added a style class with id/type/name attributes, to ease working with documents where style ids are radically different from style names.

I could not wrap my head around your testing framework, so I added basic nose tests in the `notes_tests` folder.


cassidylaidlaw:
Any idea when this PR will be merged?

PR #79, opened 2014-07-17
Get all IDs ignoring prefix

rev112:
Hi,

DocumentPart.next_id returns the next available id value for the given document part, but it searches only for 'id' attributes without prefixes. What do you think about this change which takes into account ALL id attributes regardless of prefix?

A bit of background: I'm currently experimenting with adding hyperlinks support to python-docx which requires generating unique id attributes called 'w:id' for 'bookmarkStart' and 'bookmartEnd' tags. I believe the attached fix could help to reuse DocumentPart.next_id for this purpose.


scanny:
Fixed a while back, neglected to close it then.

Issue #94, opened 2014-09-19
Sizing Images


Alecat:
Hi,
I am working on a document generator for files that will eventually be imported into Adobe Indesign.
Picture scaling works correctly for images when viewed in Word/OpenOffice. However when imported into InDesign the images appear at their native resolution instead.

After doing some digging I found that the issue was with the sizing of the pic:spPr element. InDesign uses these to render the image, rather than the dimensions of the Inline object's extents.

I am not sure if this is an error with python-docx or Adobe's parsing.

I was able to fix my issue by passing width and height through all add_picture/new_picture methods and then relocating the picture scaling logic (from Run's add_picture method) into InlineShape's new_picture method. Both the picture and the inlineshape can then be updated with the correct size information and the file now imports correctly into InDesign.

Does this sound like an appropriate fix? Alternatively, is it possible to access the Picture object inside the InlineShape? That would then make it be possible to modify its dimensions in add_picture.

I have not tested this extensively, nor do I pretend to really understand the docx spec.  I don't know if I may have just broken other things but I can submit a pull request with changes if you want to see more clearly in code what problem I was attempting to solve.

scanny:
Hi @Alecat, @toplayer, finally found some time to address this.

I just pushed a temporary branch `fix/94` up to the repo here:
https://github.com/python-openxml/python-docx/commits/fix/94

Could you give it a try and confirm it fixes the problems you were having with scaled images during print etc.?

If you look at the commit you'll see the scope of the change is much more modest than your PR @Alecat, but I believe it gets the job done and has the advantage that it also works if one changes the size of the inline picture after it's created, for example to scale a picture in an existing document. e0d928d9

I'm inclined to do some refactoring along the lines of your PR @Alecat but wanted to get this problem confirmed fixed before digging into that.

Let me know after you've had a chance to test and I'll prepare a point release to get it into the main distribution.

Alecat:
Hi Steve, thanks for making that change. Sorry for not addressing the PR
sooner, I've been out of reliable internet access lately.

I'll see if I can grab the branch and test it soon.

Alecat:
Hi Steve, I've had a play and I think the issues are resolved by your
changes. Thanks for checking it out.

scanny:
This is fixed in version 0.8.2 released Feb 16, 2015.

PR #95, opened 2014-09-22
Scale image size on insert

Alecat:
As discussed: https://github.com/python-openxml/python-docx/issues/94

Allow for width and/or height to be passed into new_picture.
This allows for any scaling to be applied to both the picture and its
inline shape.


scanny:
This is fixed in version 0.8.2 released Feb 16, 2015. I took a somewhat different approach, but thanks very much for the PR Alecat, it was very helpful for thinking through how best to approach it :)
Fix commit is here: 395b48f0. The overall rewrite of Run.add_picture() is in the 7 commits starting here: 71c657db

PR #102, opened 2014-10-03
Add trPr to CT_Row class

Apteryks:
Add table row properties element (trPr) to `CT_Row` class.


Apteryks:
That one still needs some attention, though! :)

scanny:
Added in release v0.8.5, pushed on Feb 21, 2015.

Apteryks:
Thanks, Steve!

Issue #25, opened 2014-03-18
Restart numbering of an ordered list in document.


proofit404:
We can easily add ordered list with `document.add_paragraph(style='ListNumber')` code. But how can we restart its numbering?

downtown12:
Hi @scanny . I have a problem with workaround way you mentioned in stackoverflow.com. In order to restarting the list numbering, you said one need to locate the numbering definition of the style as well as the abstract numbering definition it points to. I tried but failed to locate these two definitions. Can you explain in detail how they can be located? Cuz I haven't found any documents about them.

scanny:
Can you link to the SO response? that would save me from taking the time to search it up.

downtown12:
Here is the link, the same as the link you answered for issue #87 : http://stackoverflow.com/questions/23446268/python-docx-how-to-restart-list-lettering/23464442#23464442

scanny:
@downtown12 I don't have time to develop step-by-step instructions for you, but if you're looking for the right direction to start looking for yourself, these might be helpful:

* The document object provides access to the numbering part for the document (numbering.xml). This call should get you there: `document.part.numbering_part`

* The `NumberingPart` object provides access to a sequence of `NumberingDefinition` objects: `numbering_part.numbering_definitions`

* The `NumberingDefinitions` class looks like where API support ends at the moment. You can get the `w:numbering` element from it though, which is the parent of all the numbering definitions: `numbering_definitions._numbering`. From there you'll need to work with lxml calls to get it's children using XPath and so on.

downtown12:
Thanks a lot @scanny .Your instructions help me connect the attributes you mentioned above with the elements of numbering.xml. They are pretty helpful.

yurac:
@downtown12, I alos use this feature almost in every one document. I see that all infrastructure is already in python-docx. Enabling it is easy - I attach a patch.
{code}
diff --git a/docx/document.py b/docx/document.py
index 655a70e..4a0ec06 100644
--- a/docx/document.py
+++ b/docx/document.py
@@ -51,6 +51,12 @@ class Document(ElementProxy):
         paragraph.add_run().add_break(WD_BREAK.PAGE)
         return paragraph
 
+    def get_new_list(self, abstractNumId):
+        """
+        Returns a new numId that references given abstractNumId
+        """
+        return self.numbering.numbering_definitions.add_num(abstractNumId, True)
+
     def add_paragraph(self, text='', style=None):
         """
[...]

PR #110, opened 2014-11-27
Added new add_list method to enhance lists support.

mohamedattahri:
Added support for multiple numbered/bullet lists with a new **add_list** api method. 

Fixes issue #25.
- **BlockItemContainer** has been improved with a new method to generate unique _numId_ values
- Created a new **ListParagraph** proxy that supports adding Paragraph instances sharing the same _numId_ values.
- **Paragraph** has a couple of new getters/setters to access _numId_ and _ilvl_

TODO:
- [ ] Tests 
- [ ] documentation

Example:

``` python
    document = Document()
    a_list = document.add_list(style='ListParagraph')
    # add_item() behaves exactly like Document.add_paragraph()
    a_list.add_item('first')
    a_list.add_item().add_run('second').bold = True

    # You can even create sub-lists to increase the indentation of sub-items
    b_list = a_list.add_list()
    b_list.add_item('one')
    b_list.add_item('two')

    # List of lists in the document
    print(len(document.lists))
```

### DIFF
--- a/docx/blkcntnr.py
+++ b/docx/blkcntnr.py
@@ -8,8 +8,11 @@ ones like structured document tags.
 
 from __future__ import absolute_import, print_function
 
+import sys
+import random
 from .shared import Parented
 from .text import Paragraph
+from .list import ListParagraph
 
 
 class BlockItemContainer(Parented):
@@ -22,6 +25,15 @@ class BlockItemContainer(Parented):
     def __init__(self, element, parent):
         super(BlockItemContainer, self).__init__(parent)
         self._element = element
+
+    def generate_numId(self):
+        """
+        Generate a unique numId value on this container.
+        """
+        while True:
+            numId = random.randint(0, 999999)
+            if not len(self._element.xpath("//w:numId[@w:val='%s']" % numId)):
+                return numId
 
     def add_paragraph(self, text='', style=None):
         """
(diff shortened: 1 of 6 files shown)


JuanSebastanCho:
How can I use this code in my project?

docx/blkcntnr.py
```diff
@@ -23,6 +26,15 @@ def __init__(self, element, parent):
         super(BlockItemContainer, self).__init__(parent)
         self._element = element
 
+    def generate_numId(self):
+        """
+        Generate a unique numId value on this container.
+        """
+        while True:
+            numId = random.randint(0, 999999)
+            if not len(self._element.xpath("//w:numId[@w:val='%s']" % numId)):
```
madphysicist: Do the quotes matter here? If so, shouldn't it be `'` for the outer, `"` for the inner?

docx/blkcntnr.py
```diff
@@ -23,6 +26,15 @@ def __init__(self, element, parent):
         super(BlockItemContainer, self).__init__(parent)
         self._element = element
 
+    def generate_numId(self):
+        """
+        Generate a unique numId value on this container.
+        """
+        while True:
+            numId = random.randint(0, 999999)
```
madphysicist: Instead of random numbers, why not just
```
from itertools import count
...
for numId in count(len(self._element.xpath('//w:numId')):
    if not self._element.xpath('//w:numId[@w="{}"]'.format(numId):
        return numId
```

PR #121, opened 2014-12-18
adds support for reading/writing Run text color

smartt:
The patch adds a 'color' attribute to a Run so that you can change the foreground text color.


scanny:
Hi Erik, this feature was added in version 0.8.3 today, Feb 19, 2015. We took a different approach to accommodate a broader API, but thanks for the PR all the same :)

The user docs on the color API are here:
http://python-docx.readthedocs.org/en/latest/user/text.html#font-color

PR #124, opened 2015-01-07
Assume inches if resolution unit not given in image TIFF metadata.

cmacattack:
Cannot assume that the resolution unit is given if the x and y resolution values are given.  Inches is probably the safest assumption.


scanny:
Fixed on develop. Will be in next release.

scanny:
Released with v0.8.4 today.

cmacattack:
Sweet! Thanks.

PR #130, opened 2015-01-12
Updated documentation for _ in Word style names.

virajkanwade:
I spent quite some time trying to debug this. Documentation mentions spaces, but even _ are replaced.


virajkanwade:
Adding in develop branch

PR #131, opened 2015-01-12
Updated documentation for _ in Word style names.

virajkanwade:
I spent quite some time trying to debug this. Documentation mentions spaces, but even _ are replaced.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,13 @@
 Release History
 ---------------
 
+0.7.6 (2014-12-14)
+++++++++++++++++++
+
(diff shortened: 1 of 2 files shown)


scanny:
Hi Viraj, I've been working on a large feature addition involving styles over the holidays. The API for style access is changing based on what I learned in the process. Styles will be accessed by name, including spaces and whatever else there is. It turns out the styleId is subject to arbitrary change as long as it's kept internally consistent in the XML (to connect a paragraph to a style for example). So despite its name it turns out to be a poor choice for use in the API.

Let me know if you want to know more :)

PR #136, opened 2015-01-31
Playing with Github

fstraw:
Added a comment so I can play with commits

### DIFF
--- a/docx/text.py
+++ b/docx/text.py
@@ -1,5 +1,5 @@
 # encoding: utf-8
-
+"""Added cray comment for testing"""
 """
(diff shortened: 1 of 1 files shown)


scanny:
Hi Brandon, it's probably better to avoid sending practice pull requests to a production repo. There are some here you can use to experiment with that won't trigger an email alert to anyone :)
https://github.com/octocat/

fstraw:
I'm sorry about that, it was an honest mistake. I was just playing around with some features and sent it to the wrong place :-/

scanny:
no worries Brandon :)

PR #140, opened 2015-02-10
Avoid triggering the "style lookup by style_id is deprecated" warning ourselves.

stevecohen42:
The API has changed to use style names rather than style IDs.  This is a great step forward, but it now generates "style lookup by style_id is deprecated" warnings if headings (with a level number) or tables (without a style) are added to the document.  This pull request fixes the automatic style name generation for headings and uses a valid style name for new tables.

### DIFF
--- a/docx/api.py
+++ b/docx/api.py
@@ -40,8 +40,8 @@ class Document(object):
         Return a heading paragraph newly added to the end of the document,
         populated with *text* and having the heading paragraph style
         determined by *level*. If *level* is 0, the style is set to
-        ``'Title'``. If *level* is 1 (or not present), ``'Heading1'`` is used.
-        Otherwise the style is set to ``'Heading{level}'``. If *level* is
+        ``'Title'``. If *level* is 1 (or not present), ``'Heading 1'`` is used.
+        Otherwise the style is set to ``'Heading {level}'``. If *level* is
         outside the range 0-9, |ValueError| is raised.
         """
         if not 0 <= level <= 9:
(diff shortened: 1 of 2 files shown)


scanny:
Good catch Steve! :)

Didn't think of that one and the tests of course completely miss that sort of thing.

I'll get it merged in here and get a point release out.

scanny:
Fixed in release v0.8.1 on Feb 10, 2015.

Thanks again for the catch Steve, I credited you the commit: b28b6cda :)

stevecohen42:
Steve,

     Thanks for the quick response.  I saw the other pull-requests sitting
there and was afraid you were not actively working on python-docx any
more.  Then I noted that you had produced the release just the day before
and...

     I am not at all certain I found all the documentation that needs to be
updated for this change.  You might just keep it in mind as you go about
the usual process of updating, etc.  I did some grepping and such, but
things still slip through.

     Thanks again for all your hard work on this project.

Cheers,

Steve

scanny:
We get a bit of a backlog on pull requests. Most of them require a lot of additional work to meet the standards for being committed. In particular, most show up without tests. But they're still valuable to keep around. Sometimes another community member can use one as a holdover while waiting for a feature or as a starting point for their own efforts. And they often serve as a useful reference point when we work on the feature in earnest.

I went back to the acceptance tests and worked my way forward on these changes, so I think we've got this change well covered. It's only in the API-level that we call the library from itself like that, to provide convenience functions, and I grepped the full project for those calls, so I feel like we've done our due diligence. If someone finds an outlier I'm sure we'll hear of it :)

PR #250, opened 2016-01-26
Support some missing tags

solarmicrobe:
Add other w:rPr tags
Add w:shd support to w:r, w:p, w:tbl, and w:tc w:*Pr tag
Add CT_Shd
Add ST_Shd
Add ST_ThemeColor
Add ST_UcharHexNumber

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 _scratch/
 Session.vim
 /.tox/
+.cache/
+.idea/
--- a/docx/mixins/PrMixin.py
+++ b/docx/mixins/PrMixin.py
@@ -1,0 +1,26 @@
+# encoding: utf-8
+
+"""
+Mixin for common cross-class variable access
+"""
+
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+
+class PrMixin(object):
+    def _get_pr_node(self):
+        for v in dir(self._element):
+            if v.endswith('Pr') and not callable(getattr(self._element, v)):
+                return getattr(self._element, v)
+        else:
+            raise AttributeError(str(type(self)) + 'has no attribute ".Pr"')
+
+    def _get_or_add_pr_node(self):
+        for v in dir(self._element):
+            if v.endswith("Pr") and v.startswith("get_or_add") and callable(getattr(self._element, v)):
+                methodToCall = getattr(self._element, v)
+                return methodToCall()
+        else:
+            raise AttributeError(str(type(self)) + 'has no attribute ".Pr"')
--- a/docx/mixins/ShdMixin.py
+++ b/docx/mixins/ShdMixin.py
@@ -1,0 +1,127 @@
+# encoding: utf-8
+
+"""
+Mixin for shared functions applying to all related objects.  To be used for rPr, pPr, tbl, and tcPr tags.
+"""
+
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+from ..shading import Shd
+from ..mixins.PrMixin import PrMixin
+
+
+class ShdMixin(PrMixin, object):
+    @property
+    def shd(self):
+        return Shd(self._element)
+
+    @property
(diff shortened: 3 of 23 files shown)


scanny:
Way too much for a single commit. Looks intriguing though. Want to describe what you've done and your approach and how it affects the calling protocol and we can take it from there?

solarmicrobe:
I started by just wanting to add w:shd to w:tcPr but to implement w:shd completely I had to implement it in w:rPr, w:pPr, w:tcPr, and w:tblPr.  As I went down that implementation I had to add CT_Shd, ST_Shd, ST_ThemeColor, and ST_UcharHexNumber. I also had to make a change to ST_HexColor to support ST_HexColorAuto as ST_HexColor is actually a union of ST_HexColorAuto and ST_HexColorRGB (which is what I moved the original ST_HexColor in to).  Then came the tests.

I used ECMA-376 Fourth Edition as I didn't see a way in the library to define a supported version to adhere to.

I've only been programming in Python for about 7 months (although a professional software enginner for 8) so all the constructive criticism the better.

scanny:
Ok, I think I'm starting to see what you're doing now.

So big thing would be to make them step by step, one commit per *atomic* piece of functionality. That would include a new unit test (*very* occasionally two are needed) and the code that makes the test pass. Generally these are in the context of a series of commits of this sort getting an acceptance test to pass. This might be different because it sounds like it's purely oxml infrastructure you're adding. I can take a closer look at that a bit later.

This atomic step commit process makes the commits easy to review, validate, and ultimately merge. If you review the commit history a bit I think you'll readily recognize the pattern I'm describing. Look for a commit that begins with `acpt:` and the several that follow it. You'll find that over and over.

Can you mention what you wanted w:shd for? The bigger picture view? There may be some work that was done in python-pptx that shares this part of the XML Schema (it may be part of the DrawingML (DML) schema shared by Word, PowerPoint, and Excel).

Generally we haven't added oxml infrastructure by itself, only in support of API calls. This is generally good agile practice, but it also allows us to skip testing the oxml infrastructure separately since it can easily be combined with the API call tests and actually produces more robust tests that way.

So it's probably worth sketching up an analysis document to describe the API calls you're looking to support and what the Word behavior is, how it maps to the XML vocabulary, etc. Those are here: https://github.com/python-openxml/python-docx/tree/master/docs/dev/analysis/features
and look like this for example: http://python-docx.readthedocs.org/en/latest/dev/analysis/features/styles/index.html

It's worth reviewing a few of them, they're not very standardized yet, but if you browse through you'll start to get the gist. The main purpose is just to work out what the intended functionality should be, especially the API calls (since those are forever once committed). When collaborating on functionality like this they bear a greater burden, needing to bridge two minds rather than just the one.

In general you want to avoid formatting changes to existing code, unless there are bits that clearly depart from PEP-8 or current codebase conventions. It's noisy in commits and everybody has different preferences, so when there's not a clear right answer it's generally easier to go with the local convention. And it's less likely to raise an unnecessary dispute when getting your code merged. Better to raise it as a question first if you think a wholesale improvement is called for and then concentrate it all in a single commit.

I'm not sure the API is right for docx.shading.Shd. Better have a look at the MS API and see how they accomplish it. Also if there's something in python-pptx establishing a precedent. Once you start to describe the feature set in your analysis page I think I'll be able to offer more clear advice on this bit. Here's a link to the MS API for a possibly related object: https://msdn.microsoft.com/EN-US/library/office/ff839330.aspx.

Here's a link to a possibly related part of the python-pptx API:
http://python-pptx.readthedocs.org/en/latest/api/dml.html#fillformat-objects

The github repo for python-pptx is: https://github.com/scanny/python-pptx

I'm not sure if I see the point of the ShdMixin, but I may well be convinced. I've toyed with the idea of using some of those but hadn't yet. Anyway, once we get some analysis down in black and white I should be able to judge more easily.

solarmicrobe:
I'll work on these suggestions over the next couple of weeks and see what I can come up with.

PR #258, opened 2016-02-04
Feature/highlight

DKWoods:
I've made the change you requested with the enumerations.  I think that's everything.


scanny:
Yep, looks good. This analysis page is ready to merge.

One thing on the Git practices. This should be a single commit named as I'd mentioned, consistent with the convention you can see in the commit history: "docs: document font.highlight_color analysis".

You can do this by squashing, which might be the most sensible if you want to retain your individual updates until you submit. What I would probably do personally is just do an amend: `git commit -v --amend` to overwrite the prior commit with the small changes, but there's some personal preference here. In either case, in the end, when you were ready to submit, you would then do a `git push -f origin feature/highlight`.

We can worry about that later if you want, I understand the Git learning curve can be a lot to swallow in a short time :) Just wanted to stay clear about what we're heading towards :)

DKWoods:
Hi Steve,

I just spent a couple of hours trying to figure out "squashing". 
Sometimes it looks like it worked, while other times it doesn't.  I 
thought I was set until I reached the point of doing the Pull Request, 
and suddenly all my individual commits are back.  I give up, at least 
for now.

It should be easier in the future.  My current theory is that squashing 
needs to be done before pushing, not after, at least in the ways I could 
find documented on the web (using rebase -i).  Your earlier e-mail about 
--amend should work.

I've now submitted an analysis document for feature/tabstops based on 
your initial feedback.

Have a good weekend.

David



On 02/04/2016 02:31 PM, Steve Canny wrote:
>
> Yep, looks good. This analysis page is ready to merge.
>
> One thing on the Git practices. This should be a single commit named 
> as I'd mentioned, consistent with the convention you can see in the 
> commit history: "docs: document font.highlight_color analysis".
>
> You can do this by squashing, which might be the most sensible if you 
> want to retain your individual updates until you submit. What I would 
> probably do personally is just do an amend: |git commit -v --amend| to 
> overwrite the prior commit with the small changes, but there's some 
> personal preference here. In either case, in the end, when you were 
> ready to submit, you would then do a |git push -f origin 
> feature/highlight|.
>
> We can worry about that later if you want, I understand the Git 
> learning curve can be a lot to swallow in a short time :) Just wanted 
> to stay clear about what we're heading towards :)
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/python-openxml/python-docx/pull/258#issuecomment-180033631>.
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

scanny:
Hi David, why don't you mention how you're using Git and I can send you some step-by-step recipes to get you going. Once you understand how it works, you'll look back and say "This is a two-second job, but I remember how puzzling it was at first." :)  I certainly was mystified for quite a while until I got the concept.

By the way, my breakthrough understanding came from reading this document; you might find it insightful as well:
https://jwiegley.github.io/git-from-the-bottom-up/

I use Git almost completely from the command line, but just let me know what tools you're using and I'll get you set up.

scanny:
@DKWoods btw, squashing definitely needs to be done before pushing (or perhaps "re-pushing" would be more apt).

So you would do it in this order:

1. Squash commits in your local repo/branch using `git rebase -i` (or other method)
2. Push this branch only up to GitHub. You'll need to use the -force option if you've now changed one or more of the commits that were already pushed GitHub `git push -f origin feature/highlight`.

PR #261, opened 2016-02-10
feature: tab stops

DKWoods:
Analysis document for tab stop definition feature


scanny:
Ok, this commit is merged into python-openxml/feature/tabstops.

DKWoods:
Steve, your last comments were made on a temporary branch which I have now deleted.  I have updated this branch based on your feedback and it now contains just the first set of acceptance tests for your review.

features/txt-parfmt-props.feature
```diff
@@ -4,6 +4,17 @@ Feature: Get or set paragraph formatting properties
...
+  @wip
+  Scenario Outline: Get tab stops object
+    Given a paragraph having <value> tab_stop elements
+     Then paragraph_format.tab_stops is a CT_TabStops object
+      
+    Examples:  values for paragraph_format.tab_stop elements 
+      | value |
+      | no    |
+      | some  |
+
```
scanny: Okay, this is still a lot of acceptance tests to stack up before any implementation, so I'm going to push through this first one just so we're not crossing objects. Then the rest should be okay in a single commit.

* This should be a Scenario: not a Scenario Outline:, because there's only one case ...
* The title should be 'Get tab stops'. object is implied in the title for all these and just adds noise to spell it out.
* the given is "Given a paragraph format", period. All paragraph format objects have a TabStops object, whether or not it contains any tab stops. And, we always start with the parent object as the given, not further up the hierarchy.
* the then clause is "Then paragraph_format.tab_stops is a TabStops object". CT_TabStops is part of the oxml/xmlchemy layer, not an API object.
* the examples go away because it's a scenario rather than a scenario outline

scanny: Okay, this acceptance test is merged in feature/tabstops on the main repo. You should rebase onto that and you can proceed with the implementation of this bit. You can submit all those commits (there might only be one), I expect it will be close enough for me to merge with any tweaks.

I'll provide feedback on the rest now, and you can submit them in commits following the ParagraphFormat.tab_stops commit(s).

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,79 @@
+Feature: Get or set TabStops
+  In order to customize tab stops
+  As a python-docx developer
+  I need a TabStops object with read/write tab_stop properties
+
```
scanny: Let's change this like so:
```cucumber
Feature: Adjust tab stops
  In order to customize tab stops
  As a developer using python-docx
  I need ways to create, read, update, and delete tab stops
```

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,79 @@
...
+
+  @wip
+  Scenario Outline: Get tab_stop objects
+    Given a paragraph having <value> tab_stop elements
+     Then paragraph_format.tab_stops contain tab_stop objects
+      
+    Examples:  values for tab_stop elements 
+      | value |
+      | some  |
```
scanny: This one is redundant. Let's drop it. I don't see how this is testing anything that's not covered by len(), iter, or indexed access.

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,79 @@
...
+  @wip
+  Scenario Outline: Count tab stop objects
+    Given a paragraph having <value> tab_stop elements
+      Then paragraph_format.tab_stops has <lenval> length
+      
+    Examples:  length values for tab_stop elements
+      | value | lenval |
+      | no    | 0      |
+      | some  | 3      |
+
```
scanny: * I like the "Count ..." phrase, I think I'll adopt that one going forward for len() :)

* the given should be a tab_stops object (direct parent), not a paragraph (great grandparent). Let's use the phrase "Given a tab_stops ..." for this and the others too.

* I think "count" works better than "value" in this case. The values should be numeric, not text, and definitely not ambiguous (as is "some"). I'd say 0, 1, and 3, since those correspond to the classic cases zero, one, and more than one. You should be able to just have a count column in the examples then since no mapping is necessary.

* the then clause should be: "Then len(tab_stops) is <count>"

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,79 @@
...
+  @wip
+  Scenario Outline: Delete a tab stop
+    Given a paragraph having <value> tab_stop elements
+     When I delete a tab_stop
+     Then paragraph_format.tab_stops has <lenval> length
+      
+    Examples:  values for tab_stop elements 
+      | value | lenval |
+      | some  | 2      |
```
scanny: I think we want something like this here:
```cucumber
Given a tab_stops object containing <count> tab stops
 When I delete a tab stop
 Then len(tab_stops) is reduced by 1
  And the deleted tab stop is no longer present in tab_stops
```

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,79 @@
...
+
+  @wip
+  Scenario Outline: Clear tab stops
+    Given a paragraph having <value> tab_stop elements
+     When I clear the tab_stops
+     Then paragraph_format.tab_stops has <lenval> length
+     
+    Examples:  values for tab_stop elements 
+      | value | lenval |
+      | some  | 0      |
```
scanny: This can become a Scenario too since there is only one set of values. So would become something like:

```cucumber
Given a tab_stops object containing 3 tab stops
 When I call tab_stops.clear()
 Then len(tab_stops) is 0
```

docs/api/enum/WdTabAlignment.rst
```diff
@@ -0,0 +1,38 @@
...
+    Left-aligned.  (deprecated)
+
+END
+    Right-aligned.  (deprecated)
+
+NUM
+    Left-aligned!  (deprecated)
+
+CLEAR
+    Clear an inherited tab stop.
```
scanny: I don't see these enumerations being referenced from the tests, so they shouldn't be part of the commit. In general, these are added on the first commit where they raise an error if they're not there.

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,79 @@
...
+
+  @wip
+  Scenario Outline: Iterate over tab stop objects
+    Given a paragraph having <value> tab_stop elements
+     Then I can iterate over the tab_stop elements
+      
+    Examples:  values for tab_stop elements 
+      | value |
+      | some  |
+
```
scanny: Now that I think of it, I'm thinking these three tests can all be combined into a single scenario. Check out the test in features/doc-access-sections.feature:
```cucumber
  Scenario: Access section in section collection
     Given a section collection containing 3 sections
      Then len(sections) is 3
       And I can iterate over the sections
       And I can access a section by index
```

I'm thinking we should model after that one.

----

In general, you should never have a Scenario Outline containing only a single example. That's what a Scenario is for.

features/txt-parfmt-props.feature
```diff
@@ -4,6 +4,17 @@ Feature: Get or set paragraph formatting properties
...
+      
+    Examples:  values for paragraph_format.tab_stop elements 
+      | value |
+      | no    |
+      | some  |
+      
+
   Scenario Outline: Get paragraph alignment
     Given a paragraph format having <align-type> alignment
      Then paragraph_format.alignment is <value>
```
scanny: Okay David, that should do it for feedback on this one. Let me know when you're ready and I'll try to be quicker the next time :)

features/txt-parfmt-props.feature
DKWoods: > In features/txt-parfmt-props.feature 
> <https://github.com/python-openxml/python-docx/pull/261#discussion_r56744916>:
>
> > @@ -4,6 +4,17 @@ Feature: Get or set paragraph formatting properties
> >    I need a ParagraphFormat object with read/write formatting properties
> >
> >
> > +  @wip
> > +  Scenario Outline: Get tab stops object
> > +    Given a paragraph having <value> tab_stop elements
> > +     Then paragraph_format.tab_stops is a CT_TabStops object
> > +
> > +    Examples:  values for paragraph_format.tab_stop elements
> > +      | value |
> > +      | no    |
> > +      | some  |
> > +
>
> Okay, this acceptance test is merged in feature/tabstops on the main 
> repo. You should rebase onto that and you can proceed with the 
> implementation of this bit. You can submit all those commits (there 
> might only be one), I expect it will be close enough for me to merge 
> with any tweaks.
>
> I'll provide feedback on the rest now, and you can submit them in 
> commits following the ParagraphFormat.tab_stops commit(s).
>

Hi Steve,

I'm back from vacation and ready to dive in, but I'm confused by what 
you mean here by "the implementation of this bit."

I thought the process was to get a fairly comprehensive set of 
acceptance tests in place, then do a series of unit test - 
implementation pairings.  Am I supposed to do acceptance test - unit 
test - implementation of each step as we go?  If not, I'm not clear what 
is needed under "implementation" for this first acceptance test.  It 
looks like it fails at the right spot pre-implementation to me, so I 
must be missing something.

David

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

features/tab-access-tabs.feature
DKWoods: Structurally, there is no difference between 1 and 3 elements for tab_stops.  I can modify my sample document if you want, but it currently contains 2 paragraphs, one with no tab stops and one with 3 tab stops.  In my opinion, 2 tests should be adequate.

docs/api/enum/WdTabAlignment.rst
DKWoods: True.  I haven't gotten to that level of tests yet.  Sorry, I got carried away.

features/tab-access-tabs.feature
DKWoods: I'd like to test len() for 0-length tab stops, but cannot iterate or access in that case.  I'd prefer to leave len() as a separate scenario outline.

scanny: Generally I'd leave that sort of thing to the unit test, but I'm okay with it if you want.

features/tab-access-tabs.feature
scanny: I'm okay with that, but the reason I mention is that the test (or test reader) shouldn't require knowledge of the internal structure. So the classic case of 0, 1, and more than one would satisfy an experienced reader who didn't have (and may not want) that knowledge of the internals. That plus you want the implementation to have flexibility to change without requiring a change in the test.

features/txt-parfmt-props.feature
scanny: Hi David, what I meant by "the implementation of this bit" was the unit-test/implementation pairing(s) that when complete allow the acceptance test I merged to pass.

Then the rest of these acceptance tests can probably come as a group; I'm assuming it compresses to four or so rather than the original nine.

It's a judgement call how many acceptance tests ahead you develop. There's certainly nothing wrong with doing them onesy-twosy at a time with implementation in-between. The problem with stacking up too many is you're getting ahead of your understanding of the implementation and raising the risk you need to go back and change something based on what you learn.

But in this case, four or so is okay, once the actual access of a TabStops object from ParagraphFormat is separated out. These would all be tests of the collection behaviors of TabStops, so it's a coherent set.

Let me know if you need more.

scanny:
@DKWoods - I rewrote the `add ParagraphFormat.tab_stops` commit and merged it here:
https://github.com/python-openxml/python-docx/commit/1153e5bb4472d2eb2c4cfb938106ed2adb939375

You'll see it's an entirely different approach:

* There are two unit tests for ParagraphFormat.tab_stops. This is because it behaves differently on the first call than on subsequent calls (for the same ParagraphFormat object). It's a @lazyproperty, which means it only creates the TabStops object the first time it's called, then holds on to it for subsequent access.
* The main test uses mocks. A mock is a "fake" object that's not really a TabStops object in this case. The mock library provides the basic functionality and a small wrapper in tests/unitutil/mock.py makes them easier to use. You probably want to do some research if you want to understand more about how they work. Here's a good start: https://en.wikipedia.org/wiki/Mock_object
* The main reason to use a mock in this case is isolation. There's no reason for the TabStops object to actually do anything, but you can tell it's created in the right circumstances and using the right call.
* There's no implementation in the oxml layer, because there doesn't need to be to get the test to pass. All that bit comes as we get further down in the next steps.

Let me know if you need more. The mocks don't show up as much once we get to the next level because most of the properties on an object use actual XML (via cxml) to see that everything works all the way through the oxml layer.

DKWoods:
Steve,

I've got to admit, it would have been a stretch for me to come up with 
that.  Thanks.  On to the next step.

David



On 04/10/2016 08:46 PM, Steve Canny wrote:
>
> @DKWoods <https://github.com/DKWoods> - I rewrote the |add 
> ParagraphFormat.tab_stops| commit and merged it here:
> 1153e5b 
> <https://github.com/python-openxml/python-docx/commit/1153e5bb4472d2eb2c4cfb938106ed2adb939375>
>
> You'll see it's an entirely different approach:
>
>   * There are two unit tests for ParagraphFormat.tab_stops. This is
>     because it behaves differently on the first call than on
>     subsequent calls (for the same ParagraphFormat object). It's a
>     @lazyproperty, which means it only creates the TabStops object the
>     first time it's called, then holds on to it for subsequent access.
>   * The main test uses mocks. A mock is a "fake" object that's not
>     really a TabStops object in this case. The mock library provides
>     the basic functionality and a small wrapper in
>     tests/unitutil/mock.py makes them easier to use. You probably want
>     to do some research if you want to understand more about how they
>     work. Here's a good start: https://en.wikipedia.org/wiki/Mock_object
>   * The main reason to use a mock in this case is isolation. There's
>     no reason for the TabStops object to actually do anything, but you
>     can tell it's created in the right circumstances and using the
>     right call.
>   * There's no implementation in the oxml layer, because there doesn't
>     need to be to get the test to pass. All that bit comes as we get
>     further down in the next steps.
>
> Let me know if you need more. The mocks don't show up as much once we 
> get to the next level because most of the properties on an object use 
> actual XML (via cxml) to see that everything works all the way through 
> the oxml layer.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub 
> <https://github.com/python-openxml/python-docx/pull/261#issuecomment-208115343>
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

features/steps/tabs.py
```diff
@@ -0,0 +1,79 @@
...
+    context.tab_stops = paragraph_format.tab_stops
+
+
+# when ====================================================
+
+@when('I add a tab_stop')
+def when_i_add_a_tab_stop(context):
+    tab_stops = context.tab_stops
+    tab = tab_stops.add_tab()
+    tab.position = Inches(1.75)
```
scanny: Why do you set tab.position separately here rather than put the position in the .add_tab() call?

features/steps/tabs.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@when('I delete a tab_stop')
+def when_i_delete_a_tab_stop(context):
+    tab_stops = context.tab_stops
+    tab = tab_stops[1]
+    tab.delete()
```
scanny: I thought the protocol for this was going to be `del tab_stops[1]`, wasn't it?

features/steps/tabs.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# then =====================================================
+
+@then('len(tab_stops) is {count}')
+def then_len_tab_stops_is_count(context, count):
+    tab_stops = context.tab_stops
+    len_tab_stops = len(tab_stops)
+    int_count = int(count)
+    assert len_tab_stops is int_count
```
scanny: You can just make this `assert len(tab_stops) == int(count)`. It's shorter so quicker to understand, plus it mirrors the 'then' statement.

features/steps/tabs.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@then('I can iterate over the tab_stop elements')
+def then_I_can_iterate_over_the_tab_stop_elements(context):
+    ts = context.tab_stops
+    tab_stops = [t for t in ts]
+    assert len(tab_stops) > 0
+    assert all(isinstance(t, TabStop) for t in tab_stops)
```
scanny: I think you'll want to assert len(tab_stops) == 3 or whatever the expected length is. That would catch more problems (like exiting on first iteration). While you're in there I'd condense the first two lines to `tab_stops = [t for t in context.tab_stops]`.

features/tab-access-tabs.feature
```diff
@@ -0,0 +1,49 @@
...
+    Given a tab_stops having 3 tab stops
+     Then I can iterate over the tab_stop elements
+      And I can access a tab_stop by index
+
+
+  @wip
+  Scenario Outline: Add a tab stop
+    Given a tab_stops having <count> tab stops
+     When I add a tab_stop
+     Then len(tab_stops) is <lenval>
```
scanny: I think on this one we need another "Then" step (actually would start with "And") to test that they're ordered by increasing distance, don't we?

scanny:
Ok, I left some comments, I think we can turn this one around quickly, no reason not to go ahead and start on the implementation whenever you're ready. I can probably merge this tomorrow if you can turn it around that quickly.

Apologies for missing the commit. I looked in junk mail and everywhere but couldn't find any notification. Probably best to just drop me a quick email on your commits until we see the notifications working again :)

features/steps/tabs.py
DKWoods: Yes, it was, but that doesn't WORK.  I'm not quite sure why, but I suspect it's some subtlety of oxml.  tab_stops[1].delete() works where del tab_stops[1] doesn't.  But I'll try to add a __del__() method to the TabStop object.

DKWoods: Not sure why, but adding a __del__() method to the TabStop object caused a lot of trouble.  It killed the ability to add tabs.  How important is this, syntactically?

features/steps/tabs.py
DKWoods: I was thinking more generically, that you can iterate over any set of tab stop elements that have more than zero elements.  I can change it to 3, though.

DKWoods:
Hi Steve,

I can't get this one turned around today.  Too many other commitments.  
There are two issues left for me to resolve:

1.  the "del tab_stops[1]" vs. "tab_stops[1].delete()" issue.  Why the 
second method works while the first doesn't, and why I can't reconcile 
the two with a __del__() method, is a puzzle at the moment.

2.  Keeping tab_stop objects in order is proving more challenging than 
it ought to be.

I *should* have time to work on this tomorrow.

David




On 04/19/2016 12:52 AM, Steve Canny wrote:
>
> Ok, I left some comments, I think we can turn this one around quickly, 
> no reason not to go ahead and start on the implementation whenever 
> you're ready. I can probably merge this tomorrow if you can turn it 
> around that quickly.
>
> Apologies for missing the commit. I looked in junk mail and everywhere 
> but couldn't find any notification. Probably best to just drop me a 
> quick email on your commits until we see the notifications working 
> again :)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub 
> <https://github.com/python-openxml/python-docx/pull/261#issuecomment-211740338>
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

features/steps/tabs.py
scanny: Okay, I did a little more research on the `del` statement and I definitely love it a lot less now. I was under the mistaken impression that `del` just directly called the `__del__()` method of the object, but that's not how it works it turns out :)

I'd say let's go with .remove() instead (analogous to the list method of the same name):

```python
tab_stops.remove(tab_stops[1])
```

This may actually be more useful, because it takes a TabStop object reference rather than an index, so you could use it in a for loop for example:

```python
for tab_stop in tab_stops:
    if i_dont_like_this(tab_stop):
        tab_stops.remove(tab_stop)
```

scanny:
No worries David :)

features/steps/tabs.py
DKWoods: Steve,

I'd like to implement both tab_stops.remove(tab) and tab.remove().

Do I need to reflect both approaches in separate acceptance tests? 
tab_stops.remove(tab) will call tab.remove(), so it seems redundant.

And I want to make sure I understand the process from here.  My next 
step is to add implementation for only ONE of my acceptance tests, 
including the appropriate unit test(s), in essence removing one @wip per 
commit.  Is that correct?

David


On 04/19/2016 05:41 PM, Steve Canny wrote:
>
> In features/steps/tabs.py 
> <https://github.com/python-openxml/python-docx/pull/261#discussion_r60324914>:
>
> > +
> > +
> > +# when ====================================================
> > +
> > +@when('I add a tab_stop')
> > +def when_i_add_a_tab_stop(context):
> > +    tab_stops = context.tab_stops
> > +    tab = tab_stops.add_tab()
> > +    tab.position = Inches(1.75)
> > +
> > +
> > +@when('I delete a tab_stop')
> > +def when_i_delete_a_tab_stop(context):
> > +    tab_stops = context.tab_stops
> > +    tab = tab_stops[1]
> > +    tab.delete()
>
> Okay, I did a little more research on the |del| statement and I 
> definitely love it a lot less now. I was under the mistaken impression 
> that |del| just directly called the |__del__()| method of the object, 
> but that's not how it works it turns out :)
>
> I'd say let's go with .remove() instead (analogous to the list method 
> of the same name):
>
> tab_stops.remove(tab_stops[1])
>
> This may actually be more useful, because it takes a TabStop object 
> reference rather than an index, so you could use it in a for loop for 
> example:
>
> for  tab_stopin  tab_stops:
>      if  i_dont_like_this(tab_stop):
>          tab_stops.remove(tab_stop)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub 
> <https://github.com/python-openxml/python-docx/pull/261/files/dbfce4c2185a8718c44be6ddd260be5c5e6856ff..98ba6b5433023e9d8d3d32330d089063cad8e1bd#r60324914>
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

scanny: Can you help me understand your reasoning?

The behavior of TabStops.remove(tab_stop) is to remove tab_stop as a member of the TabStops sequence. So it seems natural to me that the call is on TabStops. After tab_stop is removed, it would no longer have a parent and so a .remove() call on it would be invalid.

So I'm not seeing what we would get by having both, and there would be additional complexity to account for if we did. Also, I'd like to reuse base Python method names whenever possible to minimize surprising behavior.

DKWoods: I can see wanting to eliminate a tab when working with a whole set of 
tab_stops, but I can also imagine working with an individual tab and 
wanting to delete it.  For example, if I'm working with an individual 
tab to change it's position, but move it to a position where another tab 
already exists, I might want to delete at the tab level.

I've got it implemented, and don't see any problematic behaviors. It's 
behaving exactly as I want it to.  What sort of problems should I look for?

I don't care if it's called delete() or remove(), as far as Python 
method names go, but we should be consistent across levels.

David

scanny: I notice there's a .clear() method on TabStop in the MS API. I'll accept a .clear() method on our TabStop object into the API if you like. That's consistent with the TabStops.clear() method, although they call that .clear_all() in the MS API. I suppose, on balance, it makes sense to rename TabStops.clear() to TabStops.clear_all(). That reduces risk of confusion with variable names that only differ by a single character (tab_stops.clear() vs. tab_stop.clear()). Then both these methods would be named the same as in the MS API, which is a consistency we've generally maintained throughout the library.

.remove() has an established Python meaning in the context of a list, so I wouldn't be wanting to choose a different name for that on the TabStops object, which is essentially a custom list.

Related to this, I did a little more research, and it turns out that the magic method that gets called on `del lst[i]` is `__delitem__()`, not `__del__()`. That makes me interested to try to make that work. The advantage of `del` is that it supports slices.

So maybe we can go back to the `del` syntax now instead of `.remove()`. If you have trouble with the implementation I'll jump in to get it working.

DKWoods: Hi Steve,

The revised feature/tabstops acceptance tests are now available for your 
review and hopefully your acceptance.

David



On 04/19/2016 05:41 PM, Steve Canny wrote:
>
> In features/steps/tabs.py 
> <https://github.com/python-openxml/python-docx/pull/261#discussion_r60324914>:
>
> > +
> > +
> > +# when ====================================================
> > +
> > +@when('I add a tab_stop')
> > +def when_i_add_a_tab_stop(context):
> > +    tab_stops = context.tab_stops
> > +    tab = tab_stops.add_tab()
> > +    tab.position = Inches(1.75)
> > +
> > +
> > +@when('I delete a tab_stop')
> > +def when_i_delete_a_tab_stop(context):
> > +    tab_stops = context.tab_stops
> > +    tab = tab_stops[1]
> > +    tab.delete()
>
> Okay, I did a little more research on the |del| statement and I 
> definitely love it a lot less now. I was under the mistaken impression 
> that |del| just directly called the |__del__()| method of the object, 
> but that's not how it works it turns out :)
>
> I'd say let's go with .remove() instead (analogous to the list method 
> of the same name):
>
> tab_stops.remove(tab_stops[1])
>
> This may actually be more useful, because it takes a TabStop object 
> reference rather than an index, so you could use it in a for loop for 
> example:
>
> for  tab_stopin  tab_stops:
>      if  i_dont_like_this(tab_stop):
>          tab_stops.remove(tab_stop)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub 
> <https://github.com/python-openxml/python-docx/pull/261/files/dbfce4c2185a8718c44be6ddd260be5c5e6856ff..98ba6b5433023e9d8d3d32330d089063cad8e1bd#r60324914>
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

features/tab-access-tabs.feature
scanny: Ok, sounds great David. I'll see if I can have a look tonight.

scanny:
Hi David, regarding how many commits to stack up, that's ultimately up to you, but I expect it will make sense to stack up at least a few. I'll merge as many as I can while I'm in there, so the main consideration is the time it might take for you to rework the stacked-up commits based on some misstep in a prior one.

Here are a few guidelines on the implementation commits:

Each implementation commit should address the first error message that appears when running the acceptance tests. So you run:

```bash
$ behave --stop features/tab-access-tabstops.feature
```

.. and it should fail with with an error related to something not being implemented. Like "no clear_all() method on TabStops" or whatever. So then you add that method. You do the unit test for it, get that to fail properly, then add the implementation for it. Once the new unit test passes, that's a commit. You've added one atomic unit of functionality and it should be pretty clear to an observer how you did it because it's fairly small. That makes validating and merging it quicker.

That commit should *also* change the error message you get from the acceptance test, or make it pass. More often the former in my experience because it seems to take two or three unit tests to implement the average acceptance test. This varies of course. The @wip flag needs to be removed on the same commit that makes that acceptance test pass.

This is what's meant by "test-driven development" (to me anyway). You let the acceptance test tell you what bit to implement next.

So if you were confident in the workflow, which I understand you might not be just yet, you might reasonably stack up all the commits it takes to implement the acceptance tests you've written. Then I'd start at the beginning, run the acceptance test, see where it fails, and look for the next commit to be that implementation. Little tweaks I'd do myself before merging, but big changes I'd probably comment and then you'd rework it. So the "stack depth" is all about how you balance the "risk" of rework (and the rebasing that entails). For myself, I'd stack up the whole implementation, but hey, I have good reason to be confident :)

Let me know if you need more :)

features/steps/tabs.py
DKWoods: Steve,

When I tried to implement __delitem__() on the TabStop object, then 
tried del tab_stops[2], which is a TabStop item, I got a Python error 
saying that the TabStops object (the parent) doesn't support item 
deletion.  When I assigned tab_stops[2] to a variable and del'ed that 
variable, the underlying list was not affected, though no error message 
was displayed.

So I'm sticking with your recommendations:

tab = tab_stops[index]

tab_stops.remove(tab)
- and -
tab.clear()

These methods both work.  As always, your mileage may vary.

David



On 04/20/2016 02:09 PM, Steve Canny wrote:
>
> In features/steps/tabs.py 
> <https://github.com/python-openxml/python-docx/pull/261#discussion_r60469364>:
>
> > +
> > +
> > +# when ====================================================
> > +
> > +@when('I add a tab_stop')
> > +def when_i_add_a_tab_stop(context):
> > +    tab_stops = context.tab_stops
> > +    tab = tab_stops.add_tab()
> > +    tab.position = Inches(1.75)
> > +
> > +
> > +@when('I delete a tab_stop')
> > +def when_i_delete_a_tab_stop(context):
> > +    tab_stops = context.tab_stops
> > +    tab = tab_stops[1]
> > +    tab.delete()
>
> I notice there's a .clear() method on TabStop in the MS API. I'll 
> accept a .clear() method on our TabStop object into the API if you 
> like. That's consistent with the TabStops.clear() method, although 
> they call that .clear_all() in the MS API. I suppose, on balance, it 
> makes sense to rename TabStops.clear() to TabStops.clear_all(). That 
> reduces risk of confusion with variable names that only differ by a 
> single character (tab_stops.clear() vs. tab_stop.clear()). Then both 
> these methods would be named the same as in the MS API, which is a 
> consistency we've generally maintained throughout the library.
>
> .remove() has an established Python meaning in the context of a list, 
> so I wouldn't be wanting to choose a different name for that on the 
> TabStops object, which is essentially a custom list.
>
> Related to this, I did a little more research, and it turns out that 
> the magic method that gets called on |del lst[i]| is |__delitem__()|, 
> not |__del__()|. That makes me interested to try to make that work. 
> The advantage of |del| is that it supports slices.
>
> So maybe we can go back to the |del| syntax now instead of 
> |.remove()|. If you have trouble with the implementation I'll jump in 
> to get it working.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub 
> <https://github.com/python-openxml/python-docx/pull/261/files/dbfce4c2185a8718c44be6ddd260be5c5e6856ff..98ba6b5433023e9d8d3d32330d089063cad8e1bd#r60469364>
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

scanny:
Okay, just merged the acceptance tests: https://github.com/python-openxml/python-docx/commits/feature/tabstops

You'll probably want to view the diff to understand the changes I made, also you'll probably want to rebase on it to make sure you get the same error sequence (so you get the same implementations sequence).

On to the next commit here, I'll let you know how that goes :)

docx/oxml/__init__.py
```diff
@@ -190,6 +192,8 @@ def OxmlElement(nsptag_str, attrs=None, nsdecls=None):
 register_element_cls('w:pPr',             CT_PPr)
 register_element_cls('w:pStyle',          CT_String)
 register_element_cls('w:spacing',         CT_Spacing)
+register_element_cls('w:tab',             CT_TabStop)
+register_element_cls('w:tabs',            CT_TabStops)
 register_element_cls('w:widowControl',    CT_OnOff)
```
scanny: @DKWoods All these changes are merged in here: https://github.com/python-openxml/python-docx/commits/feature/tabstops

..these comments are just so you can know for next time. Check the merged commit for actual specifics.

* w:tab/CT_TabStop is not needed to pass this test. It will have to wait it's turn :) Dropped from this commit.

docx/oxml/text/parfmt.py
```diff
@@ -8,7 +8,7 @@
 from ...shared import Length
 from ..simpletypes import ST_SignedTwipsMeasure, ST_TwipsMeasure
 from ..xmlchemy import (
-    BaseOxmlElement, OptionalAttribute, RequiredAttribute, ZeroOrOne
+    BaseOxmlElement, OptionalAttribute, RequiredAttribute, ZeroOrOne, OneOrMore
 )
```
scanny: * imports are maintained in alphabetical order

docx/oxml/text/parfmt.py
```diff
@@ -311,3 +312,17 @@ class CT_Spacing(BaseOxmlElement):
...
     line = OptionalAttribute('w:line', ST_SignedTwipsMeasure)
     lineRule = OptionalAttribute('w:lineRule', WD_LINE_SPACING)
+
+
+class CT_TabStop(BaseOxmlElement):
+    """
+    ``<w:tab>`` element, specifying tab stop values.
+    """
+    position = RequiredAttribute('w:pos', ST_SignedTwipsMeasure)
+
```
scanny: CT_TabStop is not needed to pass this test. Dropped from this commit.

docx/text/tabstop.py
```diff
@@ -13,5 +13,14 @@
...
     """
+
+    __slots__ = ()
+
+    @property
+    def position(self):
+        """
+        The position setting for this tab stop.
+        """
+        return self._element.position
```
scanny: * TabStop is not a base class. That would be something like _BaseTabStop if for some reason we had like five flavors of tab stop and they all were subclasses of _BaseTabStop.
* I prefer to reserve the term "element" in docstrings this close to the oxml layer to refer to an XML element.
* TabStop.position is not needed to pass this test. Dropped from this commit.

docx/text/tabstops.py
```diff
@@ -18,3 +18,16 @@ class TabStops(ElementProxy):
...
+    def __len__(self):
+        tabs = self._get_tabs()
+        return len(tabs.tab_lst)
+
+    def _get_tabs(self):
+        """
+        Get or Create tabs list as needed.
+        """
+        tabs = self._element.get_or_add_tabs()
+        return tabs
```
scanny: See the merged version for this. The general principle is that the API class (like TabStops) calls the oxml layer for whatever it needs in the XML domain. Most of that is provided by the meta class, so automatically available once you define the CT_x class and its ZeroOrOne's and whatever. However, if something fancier is needed, perhaps to navigate across more than one level in the XML hierarchy, additional methods are added in the oxml layer. So _get_tabs(), if it were needed, would go in CT_Tabs.

It's not needed in this case though, you can see the merged version for how that works out. Basically, the w:tabs element needs to have something in it (at least one w:tab element) or be removed. So we can't add one just to see how many are in it. We have to test for the not present condition and return zero in that case.

tests/text/test_tabstops.py
```diff
@@ -0,0 +1,36 @@
...
+    @pytest.fixture(params=[
+        ('w:p/w:pPr',                          0),
+        ('w:p/w:pPr/w:tabs/w:tab{w:pos=2880}', 1),
+    ])
+    def len_fixture(self, request):
+        tab_stops_cxml, expected_value = request.param
+        tab_stops = ParagraphFormat(element(tab_stops_cxml)).tab_stops
+        return tab_stops, expected_value
+
+    # fixture components ---------------------------------------------
```
scanny: A couple things here. First, you should never generally have to involve the parent object (ParagraphFormat in this case) when unit testing one of its children. So while ParagraphFormat is used in tests/text/test_parfmt.py for accessing the tab_stops object, it shouldn't appear for doing anything *with* a TabStops object. Likely it wouldn't appear in this file at all.

So the XML starts with the element proxied by TabStops, which in this case is w:pPr.

Related to this, you should always be able to get your unit object by calling it's constructor. Having to use ParagraphFormat.tab_stops is a code smell because it relies on the proper functioning of that method (so the test isn't isolated).

See the merged version for more details.

scanny:
Btw, the next acceptance test error now is: TypeError: 'TabStops' object is not iterable

You should get the same one when you run behave, just wanted to make sure we were in sync :)

On to iteration :)

docx/text/tabstops.py
DKWoods: I see you are putting the the TabStop class in the same file as the TabStops class, where I had them in separate files.  Is this how you want it going forward, or will we separate these classes into different files at a later time?

scanny: I've gone back and forth on that over the course of developing the library, whether to put the collection in the same file as the item it contains or whether to put the collection in the file of the object that "owns" the collection. I've pretty much settled on putting them both in the same file, so you'll see that pattern most places in the library. `docx.section` is an example of the former. `docx.style.styles` is a counter-example.

Anyway, in this case, let's locate them both in the same file until there's some compelling reason for them to be separate. I'm expecting the total code in that module to come in well under 1000 lines (which is an informal limit for this library) and I expect TabStops by itself would be pretty short.

docx/text/tabstops.py
```diff
@@ -21,6 +21,15 @@ class TabStops(ElementProxy):
...
 
+    def __getitem__(self, key):
+        """
+        Enables dictionary-style access by index.
+        """
+        tabs = self._element.tabs
+        if tabs is None:
+            return None
+        return TabStop(tabs[key])
+
```
scanny: A couple problems here:

1. The docstring indicates "dictionary-style". I think we're talking about "list-style" behavior here. Along those same lines, I'd prefer `idx` rather than `key` to impress on the reader we expect a non-negative integer here.

2. `tabs` is only the parent element, it does not have sequence behaviors (despite the naming :). So the `w:tab` element in question here comes from `tabs.tab_lst[key]`, not `tabs[key]`. I'd put that access on a separate line just to keep things easy reading. Also, it produces a marginally more useful error message when someone chooses an index out of range.

3. There are no cases here where indexed access should return `None`. It should return a tab stop or raise an exception. In an empty TabStops object for example, even tab_stops[0] is invalid. The exception should be `IndexError`.

tests/text/test_tabstops.py
```diff
@@ -34,9 +34,30 @@ def it_can_iterate_over_its_tab_stops(self, iter_fixture):
...
         assert TabStop_.call_args_list == expected_calls
 
+    def it_can_get_a_tab_stop_by_index(self, index_fixture):
+        tab_stops, expected_vals, tab_stop_, TabStop_, expected_calls = (
+            index_fixture
+        )
+        for count in range(len(tab_stops)):
+            assert tab_stops[count] is tab_stop_
+        assert TabStop_.call_args_list == expected_calls
+
```
scanny: 1. You're going to need two tests here, one for the tabs is non-empty case and one for the empty case (where it raises an exception).

2. I expect this test came from somewhere else in the code base which means I probably wrote its progenitor, but I'm inclined toward a straight-through approach here where one test case calls `tab_stops[idx]` once and only once. That makes the test method as simple and straightforward as possible. If there is merit in it being called more than once on the same XML structure, that can be implemented as multiple cases (params) in the fixture.

  ```python
  def it_can_get_a_tab_stop_by_index(self, index_fixture):
      tab_stops, idx, TabStop_, tab, tab_stop_ = index_fixture

      tab_stop = tab_stops[idx]

      TabStop_.assert_called_once_with(tab)
      assert tab_stop is tab_stop_
  ```

3. For what it may be worth, there's a non-sequential order in which I write these tests. I start with the test, then fixture comes second. For the test, I start with the action, calling `tab_stops[idx]` in this case. Next I move on to what I want to assert about that. Then I look for variables I need (flake8 built into my editor (vim) highlights those for me), and I have them come from the fixture, in the order they're needed. So middle, bottom, top order, in which the first step is easy and each step leads to the next.

  On the fixture, I start with the return statement, exactly matching the LHS of the assignment in the test. Then I work out how to get those items one by one. It's remarkably mechanical once you get the hang of it; the same patterns over and over.

tests/text/test_tabstops.py
```diff
@@ -34,9 +34,30 @@ def it_can_iterate_over_its_tab_stops(self, iter_fixture):
...
+    ])
+    def index_fixture(self, request, TabStop_, tab_stop_):
+        pPr_cxml, expected_vals = request.param
+        pPr = element(pPr_cxml)
+        tab_elms = pPr.xpath('//w:tab')
+        tab_stops = TabStops(pPr)
+        expected_calls = [call(tab) for tab in tab_elms]
+        return tab_stops, expected_vals, tab_stop_, TabStop_, expected_calls
+
+    @pytest.fixture(params=[
```
scanny: The fixture needs to be different, of course, because the items it supplies are different. But a couple remarks:

1. The values in the w:pos attributes are arbitrary. I'd use 1, 2, 3 ... so in case they show up in a test error you can identify the particular failing case easily. The return value is a tab_stop_ mock; tab_stop.position doesn't come into it yet.

2. The expected calls bit is relatively unusual. It shouldn't be needed here since there are no loops involved (if we don't put one in ourselves :). The .assert_called_once_with() method on the TabStops_ mock suffices up in the test itself.

Why don't you take another crack at this commit based on these comments and see what you can come up with.

tests/text/test_tabstops.py
DKWoods: I get an assertion error saying "Expected 'TabStop' to be called once.  Called 2 times."  I don't see how, and am so lost in Mock that I don't know how to address the problem, nor to debug it.

scanny: Did you replace this test method with the version I provided above? There's only one call (no for loop anymore), so I can't see how it would get called twice.

DKWoods: > In tests/text/test_tabstops.py 
> <https://github.com/python-openxml/python-docx/pull/261#discussion_r61119742>:
>
> > @@ -34,9 +34,30 @@ def it_can_iterate_over_its_tab_stops(self, iter_fixture):
> >          assert count == expected_count
> >          assert TabStop_.call_args_list == expected_calls
> >
> > +    def it_can_get_a_tab_stop_by_index(self, index_fixture):
> > +        tab_stops, expected_vals, tab_stop_, TabStop_, expected_calls = (
> > +            index_fixture
> > +        )
> > +        for count in range(len(tab_stops)):
> > +            assert tab_stops[count] is tab_stop_
> > +        assert TabStop_.call_args_list == expected_calls
> > +
>
> Did you replace this test method with the version I provided above? 
> There's only one call (no for loop anymore), so I can't see how it 
> would get called twice.
>

Yes, I replaced it.  But I don't know how to write the fixture.  For 
example, the fixture returns tab.  What's the value that's supposed to 
be in that parameter?  I don't know what is supposed to be passed to 
TabStop_.assert_called_once_with(tab) and I've got no clue how to figure 
it out.

I've spent all morning reading about Mock, but nothing I've found to 
read has helped even slightly.  I can't find a way in.

David

scanny: Alright, here's the fixture for it. I'll go through it line by line:

We start with this minimum known fixture, based on what we found we needed in the test. The return line matches the left-hand-side of the assignment in the test exactly, same spelling and everything:

```python
@pytest.fixture
def index_fixture(self, request):
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

It returns the right items, and is minimal everywhere else, so this is strictly mechanical. This doesn't work because it doesn't have the return values defined anywhere, so we take them one by one, starting with tab_stops. We create an element from some compact XML (cxml) and construct a TabStops object from it. That's step two done.

```python
@pytest.fixture
def index_fixture(self, request):
    pPr = element('w:pPr/w:tabs/w:tab{w:pos=0}')
    tab_stops = TabStops(pPr)
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

For idx, we'll just say 0 here, since there's only one member, that's the only valid value.

```python
@pytest.fixture
def index_fixture(self, request):
    pPr = element('w:pPr/w:tabs/w:tab{w:pos=0}')
    tab_stops = TabStops(pPr)
    idx = 0
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

TabStop_ is a class_mock that's already defined, so we bring it in as a parameter of the fixture. py.test will automatically locate it and pass it into the fixture correctly when it instantiates the fixture.

```python
@pytest.fixture
def index_fixture(self, request, TabStop_):
    pPr = element('w:pPr/w:tabs/w:tab{w:pos=0}')
    tab_stops = TabStops(pPr)
    idx = 0
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

`tab` is that w:tab grandchild element of the w:pPr, so we get that using xPath:

```python
@pytest.fixture
def index_fixture(self, request, TabStop_):
    pPr = element('w:pPr/w:tabs/w:tab{w:pos=0}')
    tab_stops = TabStops(pPr)
    idx = 0
    tab = pPr.xpath('./w:tabs/w:tab')[idx]
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

And tab_stop_ is an instance_mock that's already defined among our fixture components, so we just add that as a parameter as well:

```python
@pytest.fixture
def index_fixture(self, request, TabStop_, tab_stop_):
    pPr = element('w:pPr/w:tabs/w:tab{w:pos=0}')
    tab_stops = TabStops(pPr)
    idx = 0
    tab = pPr.xpath('./w:tabs/w:tab')[idx]
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

Now this one should run fine, but we want a few more cases, so we parameterize the fixture:

```python
@pytest.fixture(params=[
    ('w:pPr/w:tabs/w:tab{w:pos=0}',                                 0),
    ('w:pPr/w:tabs/(w:tab{w:pos=1},w:tab{w:pos=2},w:tab{w:pos=3})', 1),
    ('w:pPr/w:tabs/(w:tab{w:pos=4},w:tab{w:pos=5},w:tab{w:pos=6})', 2),
])
def index_fixture(self, request, TabStop_, tab_stop_):
    pPr_cxml, idx = request.param
    pPr = element(pPr_cxml)
    tab = pPr.xpath('./w:tabs/w:tab')[idx]
    tab_stops = TabStops(pPr)
    return tab_stops, idx, TabStop_, tab, tab_stop_
```

And Bob's your uncle :)

scanny: Btw, in response to a specific question in your prior comment:

The TabStop_ mock gets called here every time your code calls TabStop(); setting up that class mock replaces the TabStop object in the tabstops module with TabStop_. When TabStop_ gets called, it doesn't create a TabStop object, but it does remember every time it was called and with what. So when we test `TabStop_.assert_called_once_with(tab)`, we're testing to make sure our code called TabStop exactly once, and passed the w:tab element of interest as a parameter.

A mock also has a discoverable return value on .return_value. So we set TabStop_.return_value to the tab_stop_ mock (that's done in the component definition), then we can test that our code returned the return value of TabStop, as it should.

tests/text/test_tabstops.py
```diff
@@ -47,9 +47,31 @@ def it_raises_on_indexed_access_when_empty(self):
         with pytest.raises(IndexError):
             tab_stops[0]
 
+    def it_can_add_a_tab_stop(self, add_tab_fixture):
+        tab_stops, new_pos, new_len, expected_xml = add_tab_fixture
+        tab = tab_stops.add_tab(new_pos)
+        assert len(tab_stops) == new_len
+        assert isinstance(tab, TabStop)
+        # assert tab_stops.???? == expected_xml
+
```
scanny: Okay, so first thing is we don't want to use len() here, because that couples this test to the __len__() implementation, which already has its own test. The good news is we don't need it. So that line goes.

Instead, we can move the expected_xml line up there. Here's that line, then I'll explain why it works:

```python
assert tab_stops._element.xml == expected_xml
```

So the secret sauce here is that built into BaseOxmlElement is a property called `.xml` which returns the XML for that element. It's actually very handy for just taking a look at what an element's subtree looks like using a print() statement, but its core use is for just this, comparing against an expected XML snippet to make sure an operation produced the right XML. TabStops._element is the w:pPr element, so this should work just fine.

A fine point is I would use 'position' instead of 'pos'. Since .add_tab(position) is an API call, the parameter name is ours to define, and I'm thinking 'position' for that job; no reason to abbreviate and the full word is less ambiguous for an end-user learning the library. The name of the attribute `w:pos`, is *not* ours to define, and by convention we use a variable of that name to mean that attribute. So keeping them distinct makes clear to the reader we're not talking about an XML attribute here, just a float that says where we want the tab to be placed.

I would also avoid the 'new_' prefix since there's no other position to distinguish it from. It's generally good form if the main call in the test looks just like the API documentation, so in this case:

```python
tab_stop = tab_stops.add_tab_stop(position)
```

Which brings me to the method name. I think we want .add_tab_stop() here rather than add tab.

docx/text/tabstops.py
```diff
@@ -21,6 +21,16 @@ class TabStops(ElementProxy):
 
     __slots__ = ()
 
+    def __getitem__(self, idx):
+        """
+        Enables list-style access by index.
+        """
+        tabs = self._element.tabs
+        if tabs is None:
+            raise IndexError
```
scanny: Each exception we raise should include a meaningful error message to indicate to the developer what went wrong so they know where to look for the problem.

Otherwise, merged just as submitted :)
abab508

tests/text/test_tabstops.py
DKWoods: The xml comparison is failing since the TabStop doesn't yet have a position property so the correct xml cannot be generated.  Do I remove position references from the expected_xml for now or do I need to implement tab_stop.position at this stage to be able to pass the unit test?

features/tab-access-tabs.feature
```diff
@@ -21,6 +21,19 @@ Feature: Adjust tab stops
...
+      And the tab stop at index <index> has alignment <alignment>
+      And the tab stop at index <index> has leader <leader>
+     
+    Examples: tab stop positions
+      | index | position | alignment              | leader                  |
+      | 0     | 914400   | WD_TAB_ALIGNMENT.LEFT  | WD_TAB_LEADER.SPACES |
+      | 2     | 2743200  | WD_TAB_ALIGNMENT.RIGHT | WD_TAB_LEADER.DOTS   |
+      
+      
+  @wip
```
scanny: These are definitely the right capabilities to be driving next, so this seems right to me.

I would break it up a bit differently though. I'll point out specifically where in individual comments. But the commits it will drive will be the same. I expect they'd be these, not necessarily in this order:

* tab: add TabStop.position getter
* tab: add TabStop.alignment getter
* tab: add TabStop.leader getter

From there you could move on to the setters, with corresponding "Set tab stop values" acceptance test(s) to drive:

* tab: add TabStop.position setter
* tab: add TabStop.alignment setter
* tab: add TabStop.leader setter

Each of which, by the way, use the getter as part of the acceptance test, hence the importance of sequencing them this way.

And then we can go back to the write-y bits of TabStops, which should go smoothly then.

features/tab-access-tabs.feature
```diff
@@ -21,6 +21,19 @@ Feature: Adjust tab stops
...
+  Scenario Outline: Get tab stop values
+    Given a tab_stops having 3 tab stops
+     Then the tab stop at index <index> has position <position>
+      And the tab stop at index <index> has alignment <alignment>
+      And the tab stop at index <index> has leader <leader>
+     
+    Examples: tab stop positions
+      | index | position | alignment              | leader                  |
+      | 0     | 914400   | WD_TAB_ALIGNMENT.LEFT  | WD_TAB_LEADER.SPACES |
+      | 2     | 2743200  | WD_TAB_ALIGNMENT.RIGHT | WD_TAB_LEADER.DOTS   |
```
scanny: The short story on this one is to bring it down to operating on an individual tab stop, and second, to test each property separately.

By combining TabStops with TabStop here you're actually testing two things at once; TabStops.__getitem__() and TabStop.position (or .alignment, or .leader).

By combining all the properties together, you force testing combinations, which are more complex without a distinct benefit.

So I'd be seeing this divided into three separate tests like this:

```cucumber
@wip
Scenario Outline: Get tab stop position
  Given a tab stop having position <position>
   Then tab_stop.position is <value>

  Examples: tab stop positions
    | position | value  |
    | 914400   | 914400 |
    | other cases as necessary to drive out corner cases (only if necessary, otherwise this becomes a single scenario rather than a scenario outline). If there's no case where value is different than position, you can just use the single position column.
```

The similar for alignment and leader.

features/steps/tabstops.py
```diff
@@ -78,3 +79,43 @@ def then_the_tab_stops_are_sequenced_in_position_order(context):
...
+def then_the_tab_stop_at_index_index_has_leader_leader(context, index, leader):
+    tab_stops = context.tab_stops
+    new_leader = {
+        'WD_TAB_LEADER.SPACES':     None,  # WD_TAB_LEADER.SPACES,
+        'WD_TAB_LEADER.DOTS':       WD_TAB_LEADER.DOTS,
+        'WD_TAB_LEADER.DASHES':     WD_TAB_LEADER.DASHES,
+        'WD_TAB_LEADER.LINES':      WD_TAB_LEADER.LINES,
+        'WD_TAB_LEADER.HEAVY':      WD_TAB_LEADER.HEAVY,
+        'WD_TAB_LEADER.MIDDLE_DOT': WD_TAB_LEADER.MIDDLE_DOT,
+    }[leader]
```
scanny: In these situations where an enumeration is needed, I've adopted a more compact approach. The existing acceptance tests are a mixed bag on this count because I haven't gone back to update them all, but new work can use this more compact method. Here's the gist by way of example:

```cucumber
...
Then tab_stop.alignment is <value>
...

Examples: tab stop positions
  | position |
  | LEFT     |
...
```

```python
def then_tab_stop_alignment_is_value(context, value):
    expected_value = None if value == 'None' else getattr(WD_TAB_LEADER, value)
```

So we just use the string name of the enumeration member, and resolve that to the particular reference in the step code.

DKWoods:
Hi Steve,

I've committed the latest revisions to these acceptance tests.  I know you like to minimize duplicate tests, but I've proposed 2 tests for each property.  TBH, if I'd done only one for each as I'd first re-written these acceptance tests, I would have missed an important issue with the leader property.

David

scanny:
@DKWoods the unit tests didn't come through on the commits. Did you write a unit test for each property to be included in the same commit as the implementation?

DKWoods:
I'll fix that tomorrow. Trying to move too fast today with too many distractions. Sorry.


Sent from my phone. Please forgive any auto-corrects I may accidentally send. 

-------- Original message --------

scanny:
Ok, these are merged here:
https://github.com/python-openxml/python-docx/commits/feature/tabstops

Please do a diff between yours and these, there were quite a few changes. I don't have time to comment them line by line, but I think you'll see why on most when you compare.

I just commented out the "writing" scenarios on TabStops for now in a separate "fix" commit. (Note I put the new acceptance tests in a separate file.) I figured I'll do the rebasing once we're done with this feature and this way we won't have to try to remember the tests, since they seemed to be working just fine, just came in a bit early in the process.

scanny:
Okay, these four are merged here:
https://github.com/python-openxml/python-docx/commits/feature/tabstops

These went quite quickly, which is awesome :)

I'll leave a few comments on the changes I made along the way.

features/tab-tabstop-props.feature
```diff
@@ -14,6 +14,18 @@ Feature: Tab stop properties
       |    out    | -457200  |
 
 
+  @wip
+  Scenario Outline: Set tab stop position
+    Given a tab stop 0.5 inches <in-or-out> from the paragraph left edge
+     When I change the tab stop position to <position>
```
scanny: The prior acceptance tests are a bit of a mixed bag on this because I refined my approach along the way. But as I came to see the "end-user" for this library as the developer using the API, I started using language in the acceptance tests that was more directly sensible to a developer rather than a typical "application" end-user.

So these have become: 'When I assign <value> to tab_stop.position' (or whatever the property is)

This makes them easy to write because the only thing that changes is the object.property part. It also makes them easier to make sense of in the steps .py file.

tests/text/test_tabstops.py
```diff
@@ -50,17 +55,28 @@ def alignment_fixture(self, request):
...
 
     @pytest.fixture
-    def position_fixture(self, request):
+    def position_get_fixture(self, request):
         tab_stop = TabStop(element('w:tab{w:pos=720}'))
         return tab_stop, Twips(720)
 
+    @pytest.fixture(params=[
+        ('w:tab',            Twips(720),  'w:tab{w:pos=720}'),
+        ('w:tab{w:pos=360}', Twips(720),  'w:tab{w:pos=720}'),
```
scanny: I removed the first case here because it's invalid XML, so not a distinct case we need to concern ourselves with. This point is a little subtle, as you'll see in another comment here. It's *not* that we never provide invalid XML as the "before" XML for a case. It's just that this first case is not meaningfully distinct from the second case.

tests/text/test_tabstops.py
```diff
@@ -34,6 +34,11 @@ def it_knows_its_alignment(self, alignment_get_fixture):
         tab_stop, expected_value = alignment_get_fixture
         assert tab_stop.alignment == expected_value
 
+    def it_can_change_its_alignment(self, alignment_set_fixture):
+        tab_stop, alignment, expected_xml = alignment_set_fixture
+        tab_stop.alignment = alignment
```
scanny: You'll notice I changed this to `tab_stop.alignment = value`, and similarly for the two other properties.

The reason is that the test doesn't know that the value is an alignment, it can be None in many cases. Plus it keeps the tests as similar (and therefore mechanical and predictable for readers) as possible.

tests/text/test_tabstops.py
```diff
@@ -51,6 +56,27 @@ def alignment_get_fixture(self, request):
...
+         'w:tab{w:val=right}'),
+        ('w:tab{w:pos=720}',              'LEFT',
+         'w:tab{w:pos=720,w:val=left}'),
+        ('w:tab{w:pos=720}',              'RIGHT',
+         'w:tab{w:pos=720,w:val=right}'),
+        ('w:tab{w:pos=720,w:val=center}', 'LEFT',
+         'w:tab{w:pos=720,w:val=left}'),
+        ('w:tab{w:pos=720,w:val=center}', 'RIGHT',
+         'w:tab{w:pos=720,w:val=right}')
+    ])
```
scanny: I cut out a number of these cases, including all the ones that provided a `w:pos` attribute in the "before" XML. In general, the before XML should be the minimum required to demonstrate the behavior. Since the oxml layer doesn't validate on every operation, it works fine without the `w:pos` (or a pre-existing `w:val`, also a required element).

tests/text/test_tabstops.py
```diff
@@ -87,6 +92,27 @@ def leader_get_fixture(self, request):
...
+         'w:tab{w:leader=dot}'),
+        ('w:tab{w:pos=720}',                 'SPACES',
+         'w:tab{w:pos=720}'),
+        ('w:tab{w:pos=720}',                 'DOTS',
+         'w:tab{w:pos=720,w:leader=dot}'),
+        ('w:tab{w:pos=720,w:leader=dots}',   'SPACES',
+         'w:tab{w:pos=720}'),
+        ('w:tab{w:pos=720,w:leader=hyphen}', 'DOTS',
+         'w:tab{w:pos=720,w:leader=dot}')
+    ])
```
scanny: Check out how I modified these cases. Also I added one for a `None` case since that's a legitimate assignment value for an optional attribute.

tests/text/test_tabstops.py
scanny: I cut out a number of these cases, including all the ones that provided a `w:pos` attribute in the "before" XML. In general, the before XML should be the minimum required to demonstrate the behavior. Since the oxml layer doesn't validate on every operation, it works fine without the `w:pos` (or a pre-existing `w:val`, also a required element).

scanny:
I merged a commit with the add_tab_stop() + del + clear_all() back in.

scanny:
merged.

features/steps/tabs.py
DKWoods: Hi Steve,

I believe I got the "del tab_stop" mechanism working properly, but I've 
run across an issue while creating unit tests that I think reveals a 
problem that I don't know how to handle.

If I have two tabs, I can delete one or the other and get the 
appropriate expected_xml.  But if I only have ONE tab and I delete it, I 
get 'w:pPr/w:tabs' as the resulting xml rather than 'w:pPr', which I 
believe is what the API specifies.  That is, if there are NO tabs, we 
should not be getting a w:tabs element at all.

Any quick ideas on the proper remedy for this?

David

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

DKWoods: Hi Steve,

Never mind.  I got it figured out.   Unit tests to the rescue.

David



On 05/25/2016 02:28 PM, David Woods wrote:
> Hi Steve,
>
> I believe I got the "del tab_stop" mechanism working properly, but 
> I've run across an issue while creating unit tests that I think 
> reveals a problem that I don't know how to handle.
>
> If I have two tabs, I can delete one or the other and get the 
> appropriate expected_xml.  But if I only have ONE tab and I delete it, 
> I get 'w:pPr/w:tabs' as the resulting xml rather than 'w:pPr', which I 
> believe is what the API specifies.  That is, if there are NO tabs, we 
> should not be getting a w:tabs element at all.
>
> Any quick ideas on the proper remedy for this?
>
> David
>

-- 
David K. Woods, Ph.D.
Researcher, Lead Transana Developer
Wisconsin Center for Education Research
University of Wisconsin, Madison
http://www.transana.org

scanny:
Okay, these are merged here:
https://github.com/python-openxml/python-docx/commits/feature/tabstops

I added an additional test on TabStops.__delitem__() to account for exception conditions.
I also groomed the commit history so any little places we changed our mind along the way are updated in the source commit.

Want to take a crack at the elaborated TabStop.position setter that does a delete+insert_in_order?

I think once we have that we're pretty much ready to call this one done.

DKWoods:
Hi Steve,

I've checked in code that allows us to change Tab_Stop.position while 
maintaining position order in TabStops.

I had to alter some existing acceptance and unit tests that assumed a 
persistent TabStop reference, as the position change using the 
add-and-delete method broke them.  Modernizing the tests seemed like a 
better approach than trying to find a hack that accomplished the goal 
without invalidating the existing tests.

David

scanny:
yep, absolutely, that's the right choice, nothing sacred about tests (unlike API) :)

I should have time to merge it in this evening.

Have you had a chance to try it out with any of your typical use cases, now that it's done, and see how well it suits actual usage?

DKWoods:
Hi Steve,

I now have this version integrated into my file parser test tool, 
although not the full Transana application.  My use case is a little bit 
limited, in that my system only supports left-aligned tabs and does not 
support leader characters at all.  But in my tests, tabs work perfectly 
for import and export between Word and my wxPython wxRichTextCtrl-based 
editor.

David

scanny:
Sounds great David, thanks :)

I'm sure we'll hear about it soon enough if someone has a use case we haven't thought of :)

PR #267, opened 2016-02-24
DOC: Updated style documentation to reflect current code.

### LINKED ISSUE #266: Running code on doc intro page raises warning
After fixing the issue with the missing `recordset` variable, I still get a warning when running in Anaconda Py3 on Red Hat 6.5, python-docx version 0.8.5 installed via pip:

```
/home/jfoxrabi/miniconda3/lib/python3.5/site-packages/docx/styles/styles.py:54: UserWarning: style lookup by style_id is deprecated. Use style name as key instead.
  warn(msg, UserWarning)
```

The fact that the intro code is not perfect is not so bad in and of itself, but it would be nice if the warning was issued with `stacklevel=2`.

madphysicist:
Introduction demo now runs with direct copy-and-paste.
Updated warning to `stacklevel=2`.

Closes #266

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+build/
 .coverage
 /dist/
 /docs/.build/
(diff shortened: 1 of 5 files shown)


scanny:
Fixed in release v0.8.7 circa Aug 18 2018.

PR #268, opened 2016-02-25
Remove hard-coded zoom.  Fixes #254

### LINKED ISSUE #254: Default docx zoom = 203%
In python-docx/docx/templates/default.docx, when it is unzipped, line 2 of the file word/settings.xml is 

`<w:zoom w:val="bestFit" w:percent="203"/>`

This means the default template opens in MS Word at 203% zoom, which is obnoxious for testing documents created with python-docx.

While I have changed my own template to read w:percent="100", maintainers might consider setting this to a more reasonable value.

vwbusguy:
Removes the hard-coded zoom, so generated document uses "bestFit" instead.  This resolves the issue reported in #254 .


scanny:
Fixed in release v0.8.7 circa Aug 18 2018.

PR #274, opened 2016-03-08
added a non distructive method to replace text inside a run

biggosh:
only <w:t> <w:br> <w:cr> <w:tab> elements are affected by this new method. So no other side effects are produced

by this new method there is a simple and safe method to replace text inside a run

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,11 @@
 *python-docx* is a Python library for creating and updating Microsoft Word
 (.docx) files.
 
-More information is available in the `python-docx documentation`_.
+*python-docx-template* is based on python-docx with extensions to manage tables (add, delete, copy rows), to replace text in paragraphs.
+
+More functionalities will be added.
+
+More information is available for the `python-docx`_.
 
 .. _`python-docx documentation`:
    https://python-docx.readthedocs.org/en/latest/
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,4 @@
 
 More functionalities will be added.
 
-More information is available for the `python-docx`_.
-
-.. _`python-docx documentation`:
-   https://python-docx.readthedocs.org/en/latest/
+More information is available for the `python-docx` at https://python-docx.readthedocs.org/en/latest/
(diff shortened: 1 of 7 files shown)


biggosh:
I've just pushed the replace_text method I've written for paragraphs. The method searches for text over all runs in the paragraph, even if the text is splitted over more runs. It replaces all occurrences in the paragraph. The style is the one of the first run containing the prefix of old text.

scanny:
Hi @biggosh, did you close this to replace it with something else or something? I'd like to leave it open otherwise, so it doesn't get lost in the ones we've merged in and so on :)

biggosh:
Hi @scanny. I closed this pull request because I haven't received any feedback. So I understand you are not interested in features I've implemented probably because I didn't follow all the protocol you request for extensions I've implemented or because you don't like my coding style. So let me know how to proceed. You can find my work in my repository (biggosh on github). I've extender your library with the possibility to manage tables template, adding row to an existing one in a specific position, copy a row, replace text inside a cell. I'm waiting for your feedback and instruction on how to proceed. Thanks

ps. if you prefer you can contact me directly on my email address.

scanny:
Hi @biggosh, you're right that we would need a certain amount of protocol in order to incorporate your changes directly, and I understand that you don't have time for that right now, that's perfectly fine with us, but that doesn't mean your work is not a contribution. I'd like to leave the pull request open if it's okay with you. Often other folks need to do something similar and will pull something from what you've done. Also when we do get around to formally including these features we'll definitely study and learn from what you've done.

If in future you do want to go through the enhancement proposal process where we agree on the API and the tests, etc., we will welcome it :)

In the meantime, if you're willing, I'd like to reopen this pull so we can make it easier for folks to find.

biggosh:
OK for reopening.

lieuzhenghong:
This is great. @scanny, what needs to be done so that we can have it in the base branch? Can I be of help?

rasyidpurnama (approved):

rasyidpurnama (approved):

biggosh:
why the modifications (like mine replace_text) are not yet been merged?

what's missing?

thx

PR #291, opened 2016-04-20
Header and Footer

eupharis:


docx/header.py
```diff
@@ -0,0 +1,12 @@
...
+    def text(self):
+        text = ''
+        for p in self.paragraphs:
+            text += p.text + '\n'
+
+        return text
+
+    @text.setter
+    def text(self, text):
+        raise NotImplementedError('todo')
```
scanny: Okay, I just revised this commit and merged it in: https://github.com/python-openxml/python-docx/commits/feature/header

Since it's your first one of these I'm thinking it's faster that way, we can maintain momentum and I'll just note the whys and wherefores here so you can pick it up as you go :)

You'll probably want to rebase and force push after you've digested the comments so you're building on the main branch, but as long as your commits are individually mergeable without error there's some discretion on all that. I'm going to be cherry picking them and generally there are always revisions in the process. I'll leave the rebasing decision to you.

I do recommend you study the diff and understand why each change was made because they'll come up over and over. The general patterns are remarkably consistent on this project, probably because everything is just a fancy way of reading and writing to an XML file :)

----

* All classes, functions, and methods get a full docstring at their first appearance. Follow the conventions of the rest of the library. Everyone's got a favorite; but consistency is the objective here.
* Header should inherit from ElementProxy because it will wrap the `<w:header>` element. More on that later.
* These two methods are unsupported by tests and not yet indicated by the acceptance test errors. So they come out for now. They'll get their turn in due course :)

docx/section.py
```diff
@@ -42,6 +42,10 @@ def __init__(self, sectPr):
         self._sectPr = sectPr
 
     @property
+    def header(self):
+        return "I am not a header"
+
+    @property
```
scanny: * This code is unsupported by a unit test, so doesn't belong here yet. Also, it fools the acceptance test to thinking you're further along than you are. So this bit is out for now

features/sct-section-header.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Header stuff
+  Scenario: Basic
+    Given a section
+     Then section.header is a Header object
+
+    Given a header having a definition
+     Then header.is_linked_to_previous is False
+
+    Given a header having a definition
+     Then header.text is the text contained in the header
```
scanny: * The "Feature:" description needs to be more descriptive. There's a pattern to it, you can look at the other feature files to see a bunch of examples.
* There's vertical spacing between the Feature description and the Scenarios (2 blank lines). Check out the others for examples. Basically each section in the .feature files is separated by 2 blank lines.
* Each scenario (Given: block) has its own title. It should also be descriptive. Check out some of the other feature files.
* Scenarios are separated by two blank lines
* I've renamed this file `sct-access-header.feature` This will eventually hold the .first_page_header, .footer, etc. scenarios when we get to them.
* We should put the other two in a separate file named `hdr-props.feature`. A general rule of thumb is scenarios that start with 'Given an x ...' should go in a file that starts with the three-letter abbreviation for x. But lets leave them out for now and stick to this first one.
* Acceptance tests that don't pass yet need the `@wip` decorator. This allows them to be skipped during the full test run while their implementation is work in progress. A flag on the command line tells behave whether to skip them or not, something like '--tags=-wip'. You run them when you're looking for the next error and skip them when regression testing before merge.

features/steps/header.py
```diff
@@ -0,0 +1,45 @@
...
+
+from helpers import test_docx
+
+
+# given ===================================================
+
+@given('a section')
+def given_a_section(context):
+    document = Document(test_docx('doc-default'))
+    context.section = document.sections[0]
```
scanny: * This one should go in the section.py steps file. The 'Then section.header is...' one below too.
* I've left out the other two for now. I think we'll have our hands full with this one for half-a-dozen commits believe it or not. It will drive implementation of HeaderPart and a number of other dependencies.
* Any new test files required (like 'doc-default.docx') need to be included in the commit (in features/steps/test_files/). These acceptance tests need to run to an error driven by missing implementation. That means all the imports need to work and the test file loads as well (and syntax etc.). I've changed this to use an existing test file because any section will do, so no new file this time around. New files should start with the three-letter abbreviation of the main object under test, generally the same one used by the steps file, so 'sct-' in this case and 'hdr-' for most of what we do later.
* A bit further to that point, you will need to be running these tests before submission. Firstly, you'll need to work out any errors in the test code itself. Second, you'll need the "actual" error it produces to know what to implement next; Section.header in this case because the error is 'AttributeError: 'Section' object has no attribute 'header''. You'll want to be seeing that error for yourself each time.

docx/header.py
eupharis: Sounds good!

docx/section.py
eupharis: sounds good!

features/sct-section-header.feature
eupharis: Sounds good.

Makes sense about grouping all the `Given an x...` into a separate file starting with the 3 letter abbreviation for x.

wip seems handy. sounds good!

features/steps/header.py
eupharis: Hmm. Ok. Just FYI all three tests did run for me locally and all three failed at the assert step.

`doc-default` is an existing test file. I chose it because it had a single section whereas `sct-section-props` has like 3 so I thought that might muddy the waters. But I suppose it doesn't really matter. We'll just always use `sections[0]` for these tests that aren't specifically testing multi-section functionality.

It makes sense though that we want all `sct` tests to use test files that start with `sct`. I'll stick with that convention.

...

What are next steps here? Maybe take a stab at unit tests of section.header?

`class DescribeSections` in `tests.test_section` seems like the unit tests for these acceptance tests:

```
@then('I can access a section by index')
def then_I_can_access_a_section_by_index(context):
    sections = context.sections
    for idx in range(3):
        section = sections[idx]
        assert isinstance(section, Section)


@then('I can iterate over the sections')
def then_I_can_iterate_over_the_sections(context):
    sections = context.sections
    actual_count = 0
    for section in sections:
        actual_count += 1
        assert isinstance(section, Section)
    assert actual_count == 3


@then('len(sections) is 3')
def then_len_sections_is_3(context):
    sections = context.sections
    assert len(sections) == 3, (
        'expected len(sections) of 3, got %s' % len(sections)
    )
```

is that a decent template to use?

I've definitely got some reading to do about how `@pytest.fixture` works.

scanny: Oh, right, don't know how I missed the doc-default.docx, there it is right there :)

Actually I like your thinking on the single section better than mine there. It's perfectly fine if they don't match exactly, if an existing file does the trick we're not particular :) I'm not sure I'll rebase in a change now that it's merged, but your thinking is right-headed :)

Yes, the first unit test should be on section and be entitled `it_provides_access_to_its_header()`. You'll see examples of that all around if you grep 'it_provides_access_to_' in the tests directory (or ag it as I do :).

You'll want DescribeSection rather than DescribeSections. Sections is the collection of sections in a document, Section is an individual section. It's a little further down in that same file.

The py.test fixtures are pretty cool. They let you separate out the fixture code from the actual test code, which makes things a lot more modular once you get the hang of it.

The other learning curve items in the tests are using fixture params for individual test cases, and mocks (which we'll be using soon). Then I guess the cxml bit, but that's fairly intuitive for most folks it seems, cxml stands for compact XML and it's just a compact notation that's parsed to generate the full text XML snippet.

Let me know where you're at once you've had a quick look around on these and I can provide more.

docx/header.py
```diff
@@ -8,12 +8,11 @@
...
 
-class Header(ElementProxy):
+class Header(BlockItemContainer):
     """
-    The default page header of a section.
+    Proxy for ``<w:hdr>`` element in this section, having primarily a
+    container role.
     """
-
-    __slots__ = ()
```
eupharis: I used BlockItemContainer in my previous implementation of header and it works great. And its docstring says it's supposed to be used for headers.

It gives you all the `.paragraphs` and `.add_paragraphs` and all that jazz pretty much out of the box.

It's a little bit weird because the class Header needs two separate elements in two separate files to function:
```
<w:headerReference> in document.xml
<w:header> in header1.xml (or whatever)
```

docx/oxml/header.py
```diff
@@ -0,0 +1,14 @@
...
+class CT_Hdr(BaseOxmlElement):
+    """
+    ``<w:hdr>``, the container element for the header content
+    """
+    p = ZeroOrMore('w:p', successors=())
+
+    @classmethod
+    def new(cls):
+        header_elm = OxmlElement('w:hdr')
+        return header_elm
```
eupharis: I included this in this commit just so we could talk about whether or not it's basically the right approach.. Feel free to nuke it. I have a feeling it might be a while before we get to this?

docx/section.py
```diff
@@ -37,11 +38,25 @@ class Section(object):
...
+    def header(self):
+        if not self._header:
+            # here's how I did the header element in my previous implementation:
+            # from .oxml.header import CT_Hdr
+            # header_elm = CT_Hdr.new()
+            # not sure what is appropriate here for this first unit test
+            header_elm = 'pretend element'
+            self._header = Header(header_elm, self)
+
+        return self._header
```
eupharis: used a pretend element here because i wasn't sure how to get a real element without writing a bunch more tests of the new CT_Hdr element type.

tests/test_section.py
```diff
@@ -111,6 +125,22 @@ def it_can_change_its_page_margins(self, margins_set_fixture):
...
+        )
+
+    @pytest.fixture
+    def header_(self, request):
+        return instance_mock(request, Header)
+
+    @pytest.fixture
+    def header_fixture(self, Header_, header_):
+        section = Section(element('w:sectPr'))
+        return section, Header_, header_
```
eupharis: I think I am getting the hand of these fixtures and cxml!

tests/test_section.py
```diff
@@ -61,6 +65,16 @@ def document_elm(self):
...
 
+    def it_provides_access_to_its_header(self, header_fixture):
+        section, Header_, header_ = header_fixture
+        header = section.header
+        Header_.assert_called_once_with('pretend element', section)
+
+        # access header again, dont instantiate with Header_ again, return same header
+        header = section.header
+        assert Header_.call_count == 1
+        assert header is header_
```
eupharis: maybe this should be a separate test?

docx/header.py
```diff
@@ -8,12 +8,11 @@
     absolute_import, division, print_function, unicode_literals
```
eupharis: @scanny ok here's my first stab at a unit test. I modeled it after this commit:

https://github.com/python-openxml/python-docx/commit/bddd27a895ee8d3b1f83b860f96255d434515421

The first acceptance test passes now, as does this unittest!

scanny: This is great @eupharis :)

I'll have a look this evening and move this along :)

tests/test_section.py
scanny: Okay, here's what this one should look like. I'll provide commentary after the code:

```python
    def it_provides_access_to_its_header(self, header_fixture):
        section, Header_, sectPr, header_ = header_fixture
        header = section.header
        Header_.assert_called_once_with(sectPr)
        assert header is header_
```

As you can see, it's pretty close for the first one :)

The main thing driving this structure is what the `Header` object itself is going to look like. The `Header` object will be an element proxy, but that element is actually optional, as we know. All we can tell for sure is that a `w:sectPr` element will be available and if the `headerReference` element is present, it will be a child of that `w:sectPr`. So the "sure thing", i.e. the `w:sectPr`, is what we pass to the constructor. Everything else is figured out by the `Header` object when and if it needs it. So the constructor is quite simple. You get back a `Header` object that knows what it's `w:sectPr` is. Everything else happens later.

tests/test_section.py
```diff
@@ -111,6 +125,22 @@ def it_can_change_its_page_margins(self, margins_set_fixture):
...
+    @pytest.fixture
+    def Header_(self, request, header_):
+        return class_mock(
+            request, 'docx.section.Header',
+            return_value=header_
+        )
+
+    @pytest.fixture
+    def header_(self, request):
+        return instance_mock(request, Header)
```
scanny: These two (Header_ and header_) go in a separate subsection (immediately following the fixtures) entitled:

```python
# fixture components -----------------
```

If you have a look around at the other test classes you soon find an example you can cut and paste. The basic notion is that these are simple, atomic mocks that can be reused in multiple unit tests. Generally there is one fixture per test, potentially multiple components per fixture. I sense you've already worked this out, but the trailing underscore indicates it's a mock. It helps keep track of what's real and what's mock when they're put to use in the test itself.

tests/test_section.py
scanny: This will need to be just slightly different:

```python
@pytest.fixture
def header_fixture(self, Header_, header_):
    sectPr = element('w:sectPr')
    section = Section(sectPr)
    return section, Header_, sectPr, header_
````

The return values are of course in the same order as they are "unpacked" in the unit test. I order them in the unit test (which I write first generally) in the order they are used. Then I copy that down to the return statement, and create them in the order they appear in the return statement, satisfying prerequisites as necessary. You seem to have picked that up, so just in case you were wondering why things showed up that way :)  I generally find it useful to be as consistent and mechanical as possible; it speeds the development process (like conforming to a kata) and of course makes it easier to follow months later when you look at it again :)

File: tests/test_section.py [hunk 12]
```diff
@@ -11,8 +11,12 @@
 from docx.enum.section import WD_ORIENT, WD_SECTION
 from docx.section import Section, Sections
 from docx.shared import Inches
+from docx.header import Header
 
 from .unitutil.cxml import element, xml
+from .unitutil.mock import (
+    instance_mock, class_mock
+)
```
scanny: This can all go on one line. I only enclose them in parens when they overflow a 78 character line.

docx/section.py
scanny: No need for a class variable here. This will just look like this:

```python
@lazyproperty
def header(self):
    """
    Return the |Header| object representing the default header for this
    section. A |Header| object is always returned, whether such a header is
    present or not. The header itself is added, updated, or removed using the
    returned object.
    """
    return Header(self._sectPr)
```

A few remarks:

* I order methods alphabetically within the class; that saves a full scan when looking for a particular one later.
* Every method gets its full docstring on creation. Same with classes. One never seems to get back to it later :)
* You'll need a Header class definition for this to pass. Which means we'll need a new file docx/header.py.

scanny: Oops, missed the header.py further up! :)

docx/header.py
```diff
@@ -8,12 +8,11 @@
...
-from .shared import ElementProxy
+from .blkcntnr import BlockItemContainer
 
 
-class Header(ElementProxy):
+class Header(BlockItemContainer):
     """
-    The default page header of a section.
+    Proxy for ``<w:hdr>`` element in this section, having primarily a
+    container role.
```
scanny: I'm not sure yet we can directly inherit from `BlockItemContainer`. Let's leave it as `object` for now. I suspect we're going to need something like `Header.body` where the body is actually the block item container. The reason is this object has to intercept calls that need it to add the header part automatically and so forth. Anyway, we don't need block item functionality just yet, so let's just push that inheritance decision off until we get to needing it :)

docx/oxml/header.py
scanny: We don't need this yet, so we should leave it out for now. But as far as the actual structure of it, if that's what you're asking, I couldn't be sure yet. I don't see us needing the `new()` classmethod, but I haven't really thought about it.

Here is something that might give an idea though, about when we create one of these:

* False is assigned to header.is_linked_to_previous
* header looks and sees whether it has a headerReference (or whatever that element is) of the right kind. Let's say it doesn't find one, which means it needs to create that header part.
* So first thing is it needs to create a new header part, then it needs to create a headerReference element to "link" to it with an rId.
* So it gets its part. That's probably a DocumentPart object, we won't care. It could be a glossary part can also have headers I don't know. All part and relationship management is done by another part. So we delegate the adding a header part to that part (definitely a DocumentPart in our case, we just won't hardwire that into the call). So the header code goes kind of like:

```python
part = self.part  # more on this later, but you can find it around in the code in places
# I'm not sure about this signature, but the gist of it will be
new_header_rId = part.add_header()
self._sectPr.add_header_reference(DEFAULT_HEADER, new_header_rId)
```

And that's all we do. We don't actually store a reference to the header, we only get that if we need it by referring to the XML and navigating the relationship etc. The XML stores the state and we don't retain a redundant copy of that state (because we don't want the job of keeping them synchronized).

Anyway, somewhere in all that a new `w:hdr` object gets created. I'm just not sure how yet. We should leave that decision until we get closer to it, probably until we're actually on the commit where we *have* to do it :) (that's the agile way :)

scanny:
Okay, I've added my comments on that commit. Want to have another go and I think we'll have this commit ready to merge :)

docx/oxml/header.py
eupharis: I like the idea of the XML being the sole source of truth about state. Looking forward to seeing how that works.

docx/header.py
eupharis: @scanny ok i think this commit is good to go. unittest and acceptance test both pass.

I went with ElementProxy instead of object for Header's base class just so that I could pass sectPr without writing some custom init method.

Class methods should be in alphabetical order. I didn't worry about tests though. It seems like those aren't alphabetical currently.

`lazyproperty` is cool! I have done that a bunch but I never thought to abstract it out like that. Super handy.

What's next? The end goal right now is "Able to load a document with header text and access that text" right?

Lots of steps to get there.

docx/header.py
```diff
@@ -7,13 +7,13 @@
...
 )
-
 from .shared import ElementProxy
 
 
 class Header(ElementProxy):
     """
-    The default page header of a section.
+    Proxy for ``<w:hdr>`` element in this section, having primarily a
+    container role.
```
scanny: This commit is merged now, these comments are just for reference for future commits :)
https://github.com/python-openxml/python-docx/commits/feature/header

I left this change out. The docstring is for the library end-user (developer using python-docx) because it appears in the API documentation. So the details of what's being proxied wouldn't generally be a concern for them. I know I've used the "container role" phrase elsewhere, but I'm thinking if we want to elaborate we'd want to do so with specifics that would be helpful to a developer learning how to use this object. I'm sure we'll find more tidbits to add here as we go.

docx/section.py
```diff
@@ -7,6 +7,8 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 from collections import Sequence
+from .shared import lazyproperty
+from .header import Header
```
scanny: Includes are alphabetical by module name (e.g. header comes before shared). Also a convention for this library is to put a blank line between the __future__ bit and the rest of the includes, then also separate Python library module includes from "local" includes by a blank line.

docx/section.py
```diff
@@ -80,6 +83,16 @@ def gutter(self):
...
         self._sectPr.gutter = value
 
+    @lazyproperty
+    def header(self):
+        """
+        Return the |Header| object representing the default header for this
+        section. A |Header| object is always returned, whether such a header is
+        present or not. The header itself is added, updated, or removed using the
+        returned object.
+        """
```
scanny: All formatting guidance is specific to this project btw. I don't subscribe to religious views about that sort of thing, but I do value the speed and sharper vision that consistency within the project brings :)

* No blank line between docstring and first method.

* Docstrings in the project (and generally everything in a .py file) are wrapped to 77 (vim: :set textwidth=77). This allows a commented-out section to pass the flake8 validation. (I have that built into my MacVim btw, pretty sure it's syntastic that runs flake8 every time I write the file.)

* flake8 should also be run as part of the tests. I have 't' set up as an alias that maps to this:

  ```bash
  :( ± alias t
  t='make clean && flake8 && py.test && behave --stop'
  ```

* '@wip' comes off the acceptance test on the first commit where it passes. This time it's only the one commit. Often it's a few or handful.

scanny:
As far as next steps are concerned, it's probably time to start a new `hdr-header-props.feature` file and start on the header properties.

In general we'll want to stay to the "getter" side of things until that's pretty well fleshed out; the acceptance tests for the "setter" side generally use the getter side methods to validate they worked.

I'm thinking the getter aspect of `Header.is_linked_to_previous` is a good place to start. So something like:

```cucumber
Scenario Outline: Get Header.is_linked_to_previous
  Given a header <having-or-no> definition 
   Then header.is_linked_to_previous is <value>

  Examples: ...
    | having-or-no | value |
    | having a     | False |
    | having no    | True  |
```

That will get us down to the next step in the process, which is accessing the right header definition elements under w:sectPr.

After that, I'm thinking the next step is `Header.body`, which is the actual block item container. Something like:

```cucumber
Scenario: Get Header.body
  Given a header having a definition 
   Then header.body is a BlockItemContainer object
```

This will need to get more sophisticated later, adding a case for a header that *doesn't* have a definition and may need to create one, but this will be a start.

Want to work in that direction and we'll see where we go?

eupharis:
@scanny ok! I just did that acpt test and pushed it. I tried to keep it to just the acceptance test and nothing else for that commit.

Tomorrow I'll dive in to the implementation and unittests.

scanny:
Okay, this one is merged here:
https://github.com/python-openxml/python-docx/commits/feature/header

Take a look at the steps. Those can be parameterized like the Cucumber steps, so only two were needed rather than four.

docx/oxml/section.py
```diff
@@ -42,12 +44,16 @@ class CT_SectPr(BaseOxmlElement):
...
     __child_sequence__ = (
-        'w:footnotePr', 'w:endnotePr', 'w:type', 'w:pgSz', 'w:pgMar',
-        'w:paperSrc', 'w:pgBorders', 'w:lnNumType', 'w:pgNumType', 'w:cols',
-        'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
+        'w:headerReference', 'w:footnotePr', 'w:endnotePr', 'w:type', 'w:pgSz',
+        'w:pgMar', 'w:paperSrc', 'w:pgBorders', 'w:lnNumType', 'w:pgNumType',
+        'w:cols', 'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
         'w:textDirection', 'w:bidi', 'w:rtlGutter', 'w:docGrid',
         'w:printerSettings', 'w:sectPrChange',
     )
```
eupharis: I added headerReference to the beginning here because that's where it appears in the analysis

eupharis:
@scanny ok unit tests added! tests and flake8 both pass.

this is definitely just a v1. I imagine there are better ways to do this implementation!

tests/test_header.py
```diff
@@ -0,0 +1,31 @@
...
+    @pytest.fixture(params=[
+        ('w:sectPr',                                    True),
+        ('w:sectPr/w:headerReference{w:type=default}',  False),
+        ('w:sectPr/w:headerReference{w:type=odd}',      True),
+    ])
+    def header_linked_previous_fixture(self, request):
+        sectPr_cxml, expected_linked_previous = request.param
+        sectPr = element(sectPr_cxml)
+        header = Header(sectPr)
+        return header, expected_linked_previous
```
scanny: * On the future includes, I generally include division as well on new modules (and when I think of it when in existing modules). I've found it too hit-or-miss to remember on first use.
* This is a test on Header, so it's DescribeHeader, not DescribeSection
* blank space between class decl and first method decl. No docstrings on these so it's a bit different from convention on library code.
* The general convention on fixtures is x_fixture, where x is exactly the property or method name. So this would be is_linked_to_previous_fixture. However, space is at a premium, so if a shorter version works for a long one like this it's generally a good idea. I'd say 'is_linked_fixture'.
* expected_linked_previous should be expected_value. That varies throughout the tests, I originally tried to match, but ended up standardizing on just expected_value.
* since sectPr is not used again in the fixture, I generally put the 'element(sectPr_cxml)' inline into the Header() call.

docx/header.py
```diff
@@ -17,3 +17,23 @@ class Header(ElementProxy):
...
+        """
+        Boolean representing whether this Header is inherited from
+        a previous section.
+        """
+        # self._element should probably be renamed self._sectPr,
+        # like it is on Section
+        if self._element.has_header_reference(self.type):
+            return False
+        else:
+            return True
```
scanny: * We'll be needing an enumeration here I think, the WdHeaderFooterIndex one. If we do need type, we'll want to store it as an instance variable because we should be able to reuse this class for any header. It will need it's own __init__() then of course that calls it's parent's __init__() with the needful; and __slots__ will need a place for the instance variable.
* We'll want to stay as general purpose as possible, so I'm thinking a method get_headerReference_of_type(type). This will come in handy later as well. It should return the matching headerReference element or None if not present.

docx/oxml/section.py
```diff
@@ -42,12 +44,16 @@ class CT_SectPr(BaseOxmlElement):
     ``<w:sectPr>`` element, the container element for section properties.
     """
     __child_sequence__ = (
-        'w:footnotePr', 'w:endnotePr', 'w:type', 'w:pgSz', 'w:pgMar',
-        'w:paperSrc', 'w:pgBorders', 'w:lnNumType', 'w:pgNumType', 'w:cols',
-        'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
+        'w:headerReference', 'w:footnotePr', 'w:endnotePr', 'w:type', 'w:pgSz',
+        'w:pgMar', 'w:paperSrc', 'w:pgBorders', 'w:lnNumType', 'w:pgNumType',
+        'w:cols', 'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
```
scanny: * the convention for these changed to _tag_seq = (..., you'll see it around in the oxml layer. Also you see on those a `del _tag_seq` statement after the element declarations. If you have a look at a few of those you'll see how to tidy this one up.
* The child tag names should appear in exactly the same sequence as they do in the XML schema for this element; they're used for sequencing new tags in the right order.

docx/oxml/section.py
```diff
@@ -119,6 +125,18 @@ def gutter(self, value):
...
+        if the corresponding ```w:headerReference``` exists in the sectPr
+        else |False|
+        """
+        HR_TYPE = '{%s}type' % NS.WML_MAIN
+        for ref in self.headerReference_lst:
+            if ref.attrib.get(HR_TYPE) == header_type:
+                return True
+
+        return False
+
```
scanny: The get_headerReference_of_type() method you write will replace this one. The True or False will just be whether it returns None or not. It will be reused in a get_or_add_headerReference_of_type() method I expect we'll need just a bit later.

Note the element name is used just as it appears in the XML, and is not converted to snake-case.

scanny:
@eupharis - Okay, I've left some comments that should give you what you need for this one. I expect we'll be able to merge it once you've had a second go, it's pretty much there :)

docx/oxml/section.py
```diff
@@ -41,22 +45,24 @@ class CT_SectPr(BaseOxmlElement):
...
     ))
     pgSz = ZeroOrOne('w:pgSz', successors=(
-        __child_sequence__[__child_sequence__.index('w:pgSz')+1:]
+        _tag_seq[_tag_seq.index('w:pgSz')+1:]
     ))
     pgMar = ZeroOrOne('w:pgMar', successors=(
-        __child_sequence__[__child_sequence__.index('w:pgMar')+1:]
+        _tag_seq[_tag_seq.index('w:pgMar')+1:]
     ))
+    del _tag_seq
```
eupharis: do you want me to list ALL the tag_seq attributes before this delete? Like on CT_Style

I think most of the attributes here in _tag_seq are OptionalAttributes. And so shouldn't be listed in this sequence thing? At least they aren't in CT_Style

eupharis:
@scanny ok I pushed fixes.

What's next? This guy?

    Scenario: Get Header.body
      Given a header having a definition 
       Then header.body is a BlockItemContainer object

docx/oxml/section.py
scanny: We distinguish here between child elements and attributes of the current element. `_tag_seq` enumerates the possible child elements, in sequence, just to provide a compact way to specify successor elements for any given child tag/element.

So ZeroOrOne() is logically ZeroOrOneChildElementHavingThisTag() whereas Optional/RequiredAttribute define an attribute on this element, rather than a child element.

As far as what goes before the `del _tag_seq`, just have a look around and make it consistent with the first couple others you find that show a similar situation. I don't remember off the top of my head what convention I've adopted. I do note that elements seem to be kind of one or the other, i.e. have children or have attributes, but I expect there are a couple cases around :)

Both ZeroOrOne and Optional reflect optionality, like minOccurs="0" or use="optional". Optionality is a bit different on attributes since the most they can appear is once where a child element can be repeating.

docx/enum/header.py
```diff
@@ -0,0 +1,39 @@
...
+        XmlMappedEnumMember(
+            'PRIMARY', 1, 'default', 'Default/odd header type.'
+        ),
+        XmlMappedEnumMember(
+            'FIRST_PAGE', 2, 'first', 'First page header type.'
+        ),
+        XmlMappedEnumMember(
+            'EVEN_PAGES', 3, 'even', 'Even page header type.'
+        ),
+    )
```
scanny: These are merged, these comments are for future reference to account for the changes I made during merge.

* This one is only used internally. As far as I know there is no public API that uses this and there won't be a `.type` property on Header, so I removed the example.
* The alias, if there is one, is the part of the MS member names that's the same for each member. In this case it's WdHeaderFooter -> WD_HEADER_FOOTER
* Enum modules get a docstring, as does every module.
* The description string for each member is generally copied directly from the MS API web page, adapted only when necessary. This one requires quite a bit of interpolation since it's used a bit differently and the MS ones are misleading and in cases just plain wrong :)
* The enumeration documentation page is generated automatically from the enumeration definition like this:

  ```python
  >>> from docx.enum.header import WD_HEADER_FOOTER as WD
  >>> print WD.__docs_rst__
```

```rst
.. _WdHeaderFooterIndex:

``WD_HEADER_FOOTER_INDEX``
==========================

alias: **WD_HEADER_FOOTER**

Specifies the type of a header or footer.

----

PRIMARY
    The header or footer appearing on all pages except when an even and/or
    first-page header/footer is defined.

FIRST_PAGE
    The header or footer appearing only on the first page of the specified
    section.

EVEN_PAGES
    The header or footer appearing on even numbered (verso) pages in the
    specified section.
```

This text is used to create the documentation page you can see in the merged commit (along with it's index entry).

docx/header.py
```diff
@@ -16,4 +18,28 @@ class Header(ElementProxy):
     The default page header of a section.
     """
 
-    __slots__ = ()
+    __slots__ = ('type')
+
+    def __init__(self, element, parent=None, type=WD_HEADER_TYPE.PRIMARY):
+        super(Header, self).__init__(element, parent=parent)
+        self.type = type
```
scanny: * All private variables have a leading underscore (`_type` rather than `type`)
* parent and type are both required, so no need to provide defaults.
* added `self._sectPr = element` for clarity, `self._element` is retained as it is provided by ElementProxy and occasionally is useful for general-purpose methods that don't have any way of knowing the specific element spelling.
* created and inherited from `_BaseHeaderFooter` since is_linked_to_previous will be general to all header/footer types.

docx/oxml/section.py
```diff
@@ -41,22 +45,24 @@ class CT_SectPr(BaseOxmlElement):
...
     ``<w:sectPr>`` element, the container element for section properties.
     """
-    __child_sequence__ = (
-        'w:footnotePr', 'w:endnotePr', 'w:type', 'w:pgSz', 'w:pgMar',
-        'w:paperSrc', 'w:pgBorders', 'w:lnNumType', 'w:pgNumType', 'w:cols',
-        'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
+    _tag_seq = (
+        'w:headerReference', 'w:footnotePr', 'w:endnotePr', 'w:type', 'w:pgSz',
+        'w:pgMar', 'w:paperSrc', 'w:pgBorders', 'w:lnNumType', 'w:pgNumType',
+        'w:cols', 'w:formProt', 'w:vAlign', 'w:noEndnote', 'w:titlePg',
```
scanny: * ZeroOrMore is alphabetically prior to ZeroOrOne (on import line)
* That import Line extends beyond 77, wrapped
* When a _tag_seq is added (or touched), all defined elements are included. Added w:footerReference. Usually this is just a legacy issue from back when I was developing that convention :)
* The _tag_seq usage is substantially more elegant now, check the merged commit. That double child_sequence bit must have been one of my earliest trials as I was developing the convention because it is butt-ugly :)

docx/oxml/section.py
```diff
@@ -102,6 +108,18 @@ def footer(self, value):
...
+        if the corresponding ```w:headerReference``` exists in the sectPr
+        else |False|
+        """
+        HR_TYPE = '{%s}type' % NS.WML_MAIN
+        for ref in self.headerReference_lst:
+            if ref.attrib.get(HR_TYPE) == WD_HEADER_TYPE.to_xml(header_type):
+                return ref
+
+        return None
+
```
scanny: Check merged commit for different approach.

scanny:
Ok, this commit is merged with changes. Please diff and read through the comments.

Yeah, I'm thinking header.body is a pretty good next step. I can't think of anything else we'd need first :)

docx/enum/header.py
eupharis: Woops! Forgot the docstring. I suspect that's not the last time that's going to happen ;)

Ok cool. Makes sense.

docx/header.py
eupharis: Sounds good!

docx/oxml/section.py
eupharis: Wow! Looks a lot better now. Cool.

docx/oxml/section.py
eupharis: Ok gotcha. I wasn't distinguishing between Child Elements and Tag Attributes. That makes sense!

Makes sense that an OptionalAttribute can appear only once.

docx/oxml/section.py
eupharis: Oh much better! That xpath looks handy.

eupharis:
Ok nice. Merged commit looks much better.

I'll get started on header.body now!

tox.ini
```diff
@@ -1,7 +1,10 @@
 #
 # Configuration for tox and pytest
 
+# first five ignores are defaults per http://flake8.readthedocs.io/en/latest/config.html#default
 [flake8]
+ignore = E123, E133, E226, E241, E242, W503, E402
+max-line-length = 77
 exclude = dist,docs,*.egg-info,.git,ref,_scratch,.tox
```
eupharis: I added this to my vimrc:
```
let g:syntastic_python_flake8_args = "--config=./tox.ini"
```
so that I catch when I'm going over 77 characters on this project.

There's a lot of 78 or 79 character long lines in the project, so flake8 doesn't actually pass with this config.

What do you think about having the flake8 config in here?

docx/header.py
```diff
@@ -34,6 +35,13 @@ def is_linked_to_previous(self):
             return True
         return False
 
+    @lazyproperty
+    def body(self):
+        """
+        BlockItemContainer instance with contents of Header
+        """
+        return BlockItemContainer(None, self)
```
eupharis: Eventually this needs to be instantiated with the `<w:hdr>` element from `/word1/header1.xml` (or whatever), not `None`.

I know we need to add this to `docx/oxml/__init__.py`:
```
from .header import CT_Hdr
register_element_cls('w:hdr',      CT_Hdr)
```
But not quite sure how we want to knit together all the xml files, the XmlParts, and the elements here...

eupharis:
@scanny Ok! I think this guy is ready for review. What's next?

tox.ini
scanny: The flake8 limit should be 79 by default. The 77 limit is just a project convention, for uncommented code, so that if you *do* comment it out it doesn't go past the 79 limit. So I don't see a way to enforce it with flake, I just use the textwidth in vim to wrap it appropriately while editing.

scanny: @eupharis - I was just looking at the features/steps/test_files/hdr-header-props.pptx. It looks like you embedded the header in a table. Is there some reason for that? I was just thinking we're probably better off for initial testing purposes if it's just as simple as possible, like the 'S1P1' in a single paragraph. Was there a specific reason you wanted to use a table?

Also, the document part is named document2.xml, which seems a little odd. How did you create the .docx file?

Also, what did S1P1 stand for? I'm thinking section 1, paragraph 1, which seems like a sensible convention except we'd need it to be S1HP1 or something, wouldn't we? To indicate whether it's a header or footer?

docx/header.py
scanny: Okay, so this one is going to need to be more sophisticated. Here's what I think is the rough logic.

* Look in the sectPr for a headerReference of the right description (type="default" I think in this case).
* If not found, return None
* If it is found, get the rId from it, that's our key to opening the part.
* call self.part.related_header_body(rId).

  This method doesn't exist yet, so it will need to be mocked in the test. A grep on 'method_mock' in the tests will find examples.

  This method basically delegates getting the right header.xml part and loading it into a HeaderBody object, to the document part (DocumentPart) object. All relationship traversal and general part finagling is the responsibility of another part (happens at the part <-> part level). docx.parts.document.DocumentPart.numbering_part is some sort of example; there are a couple other examples on that class as well.

* HeaderBody will inherit from BlockItemContainer. BlockItemContainer is only a base class, not used directly. There will at some time be a HeaderPart class as well, in docx.parts.header.py, not sure when that will come in but it will be distinct from Header and HeaderBody.
* return header_body (the return value of self.part.related_header_body(rId)).

Just as a sketch I think it might look roughly like this:

```python
headerReference = self._sectPr.headerReference_of_type(self._type)
if headerReference is None:
    return None
return self.part.related_header_body(headerReference.rId)
```

The analysis document should also be elaborated a bit to show an XML specimen of an actual header (w:hdr at root), say with a single paragraph containing 'Foobar' or whatever. The rich details are all defined by BlockItemContainer so don't need to be repeated there. Also we'll want the XML Schema bits that rule w:hdr and any other bits and pieces you notice we're likely to need. And the parts of the protocol that we're into now should be added as well if they're not there yet. It might make sense to actually assemble a second document, the first for header access (getting ahold of a header) and the second for header properties (working with one once you actually have it in hand :).

I expect by now your starting to see that the analysis document is not the idle exercise it perhaps initially appeared to be, but rather is a collection of all the things we need now to complete the implementation :) In fact, I didn't do those initially, it was only the suffering (in the form of rework) for the lack of that information that drove me to that extra first step :)

scanny:
btw, I merged (a somewhat elaborated version of) the acceptance test commit here:
https://github.com/python-openxml/python-docx/commits/feature/header

tox.ini
eupharis: I just quickly created that file with header via the Word Online interface. So it has whatever strangeness came from that! We should probably replace it with a "purer" docx with just the header.

S1HP1 sounds good to me!

eupharis: re flake8 79 limit gotcha! I'll use textwidth in vim too

docx/header.py
eupharis: Yeah! The analysis document is handy. I can add the header specimen XML and the new protocol parts.

I like the structure of that `self.part.related_header_body(headerReference.rId)`. Seems like it should work and be testable.

I'm helping give a tutorial at PyCon on Sunday so I'm going to be busy with that the next few days. And then I'll be at PyCon proper. So I probably won't make any forward progress until next Thursday at the earliest.

Are you going to PyCon?

scanny: That's great! Best of luck with the tutorial, I'd love to hear how it goes :)

I won't be going, although I will keep an eye out for next year. I've enjoyed some of the videos from prior years and my wheels are starting to spin now you mention it on how to get work to sponsor me going and maybe participating somehow as a contributor :)

eupharis:
Tutorial went surprisingly not terrible! https://www.youtube.com/watch?v=oSSnDeOXTZQ

Work sponsored me this year. Definitely the way to go :)

I rebased and revamped that unit test. I still need to make the `related_header_body` return a `HeaderBody` subclass of `BlockItemContainer` but that might be a separate commit and I think it'd be useful to talk about what I have so far.

Also the analysis is still on my radar. Next week!

scanny:
Glad to hear it Dan :)

Just let me know whenever you're ready and I'll have a look :)

eupharis:
Oh let's talk about the commit I pushed! And then I'll move on to the other parts.

docx/header.py
```diff
@@ -34,6 +35,19 @@ def is_linked_to_previous(self):
...
+        """
+        BlockItemContainer instance with contents of Header
+        """
+        headerRef = self._sectPr.get_headerReference_of_type(self._type)
+        if headerRef is None:
+            return None
+
+        # should always have an r:id or else it's a really weird headerRef...
+        rId = headerRef.attrib['{%s}id' % NS.OFC_RELATIONSHIPS]
+        return self.part.related_header_body(rId)
```
scanny: Ok, basically it looks like you haven't implemented the oxml layer yet. You shouldn't need any lxml code in this (API)  layer. So sectPr is a CT_SectPr type and is defined in docx.oxml.section. It has `w:headerReference` defined as a ZeroOrMore child, but the `w:headerReference` tag name is not registered in `docx/oxml/__init__.py`, so all headerReference elements get loaded as a regular lxml class with no custom behaviors.

In order to register it, you need a custom element class, CT_whatever, probably in docx/oxml/header.py. The class name is spelled the same as the XML complex type name. Then you can define attributes on it like id and rId.

The attribute definition would be:

```python
rId = OptionalAttribute('r:id', ST_RelationshipId)
```

Then getting the rId value is just `rId = headerReference.rId`. Note the convention is to spell the element variable name exactly as the tag name (less the namespace prefix), so headerReference rather than headerRef.

tests/test_header.py
```diff
@@ -33,3 +43,37 @@ def is_linked_fixture(self, request):
         sectPr_cxml, expected_value = request.param
         header = Header(element(sectPr_cxml), None, WD_HEADER_FOOTER.PRIMARY)
         return header, expected_value
+
+    @pytest.fixture
+    def body_fixture(self, part_prop_, block_item_):
+        # this is what i want to do but it throws an error...
+        # > sectPr = element('w:sectPr/w:headerReference{w:type=default,r:id=rId8}')
+        # < lxml.etree.XMLSyntaxError: Namespace prefix r for id on headerReference is not defined, line 2, column 50
+        # how do i fix?
```
scanny: I think this should go away once you have the `CT_HeaderThingaMaJig` element defined. Getting the error from lxml.etree... always seems to indicate an element not defined/registered yet.

If it doesn't fix it, let me know, it could be the cxml parser needs a tweak to recognize attribute namespace prefixes.

scanny:
Ah, apologies, I missed the new commit there, I've left some comments that should help you move it along :)

tests/test_header.py
eupharis: @scanny Ok cool. I updated the commit. Much cleaner using that `CT_HdrFtrRef`.

I took the name from the analysis doc. I forget why we named it that but I think we probably had a good reason. Maybe.

Adding the `CT_HdrFtrRef` didn't fix the `r:id` cxml issue. Want to take a look?

scanny:
Ok, this is merged here:
https://github.com/python-openxml/python-docx/commits/feature/header

Note I rebased in a couple commits before it to take care of the CXML parser bit you were seeing and also to make Section inherit from ElementProxy.

You might want to review the top commit, there were a few changes I made along the way you'll probably want to understand and also I added a couple important cases to the unit test, the return None case wasn't covered.

eupharis:
@scanny Ok! I reviewed the top commit. Looks good.

I just pushed those analysis updates. I think maybe we should hide the `header.body` interface, like on `document`. So the api people actually use is just `header.add_paragraph` or `header.paragraphs`, which internally uses `header._body`. What do you think?

What's next steps to code? `related_hdrftr_body`?

To add the header relationship before between `hdr` and `headerReference` I did this
```
        reltype = nsmap['r'] + '/header'
        self._parent.part.rels.add_relationship(reltype, header_part, rel_id)
```
So for this reading stage we'll have to interact with `self._parent.part.rels` and use that to get to the HeaderPart

Here's what I had for the HeaderPart before:
```
class HeaderPart(XmlPart):
    # MOSTLY COPYPASTA FROM DOCUMENT PART BELOW THIS POINT
    # TODO ABSTRACT?
    @property
    def next_id(self):
        """
        The next available positive integer id value in this document. Gaps
        in id sequence are filled. The id attribute value is unique in the
        document, without regard to the element type it appears on.
        """
        id_str_lst = self._element.xpath('//@id')
        used_ids = [int(id_str) for id_str in id_str_lst if id_str.isdigit()]
        for n in range(1, len(used_ids)+2):
            if n not in used_ids:
                return n
        ...
```
And then in the `docx/__init__.py` this guy was necessary:
```
PartFactory.part_type_for[CT.WML_HEADER] = HeaderPart
```
So in English what I understand this to do is something like:
```
For all content types in the `document.xml.rels` of type WML_HEADER, interact with the actual XML file via the HeaderPart.
```

eupharis:
Whew busy few weeks! Hope yours has been good! @scanny where are we at on this? any next steps?

scanny:
Hi Dan, apologies, this one slipped down my inbox a bit :)

It looks from the current behave error that the next step is to implement DocumentPart.related_hdrftr_body().

Hiding Header.body might possibly work. I assume the behavior you mean is that it would provide access to the paragraphs or whatever of the *effective* header, meaning this one if defined or the inherited one if not. So .body would never be None and it would add a new header on the first section if there were no other header to inherit from.

We can probably postpone a final decision until we get further along, since I'm thinking these would just amount to API shortcuts; the "long-form" calls would still need to be implemented.

On the rels stuff, most of those operations (I think all) have been factored out into docx.opc.Part. So for the situation you mention, all you need in Header should be:

```python
from docx.opc.constants import RELATIONSHIP_TYPE as RT
rId = self.part.relate_to(header_part, RT.HEADER)
```

If you look in docx.parts.styles.StylesPart.default() you can see an example of how a new part is created from scratch. SettingsPart has one too, very similarly implemented.

Let me know if you need more to go on :)

I'll wait to merge the docs updates until the next commit so you can update the page if you need to.

eupharis:
Ok! Just pushed another commit.

For some reason I find all this Part stuff way more confusing than anything else in this project. I hope this commit is going in the right direction.

`HeaderPart` feels slightly different than `NumberingPart`, `StylesPart`, and `SettingsPart` in that we can have mulitple `HeaderPart`s in the same Document and they can have arbitrary PackURIs.

With this commit, if `CT_HeaderFooter` were done the behave test might  be close to passing? Would `CT_HeaderFooter` be the next step?

scanny:
Yes, it took me a while to get the part bits worked out. Initially the part and the API object (like document) were the same object; it's only relatively recently that I made them distinct.

The thing about a part object is that it deals with packaging concerns (and interacting with other parts). This is distinct from the API object, which deals strictly with the XML *inside* the part. So that's where the boundary is.

Anytime you're dealing with an rId (relationship), it's going to involve the part, because only a part can interact with another part. Also, most of those interactions are factored out into the XmlPart or Part base classes.

Hopefully that helps clarify a little bit :)

HeaderPart *is* distinct (so far) in the way that you mention in python-docx. But it's not the first time we've encountered sets of parts. In python-pptx, slide parts (and others) are multiple. So we should have all the infrastructure in place for dealing with them built into the .opc sub-package.

I'll have a look at this commit and try to get it merged this week. I'm traveling, so it will be toward the end of the week.

docx/parts/document.py
```diff
@@ -126,7 +126,15 @@ def related_hdrftr_body(self, rId):
...
-        raise NotImplementedError
+        part = self.get_related_part(rId)
+        return part.body
+
+    def get_related_part(self, rId):
+        """ HACK this isn't strictly necessary
+        just adding it because it seems much easier to mock than
+        self.rels.related_parts
+        """
+        return self.rels.related_parts[rId]
```
scanny: Let's keep this all in one piece. Here's an example of how I mocked `related_parts` in python-pptx. `.related_parts` is available directly on the part object, no need to go out to .rels to get it :)

```python
    # in pptx.parts.presentation ------
    def related_slide(self, rId):
        """
        Return the |Slide| object for the related |SlidePart| corresponding
        to relationship key *rId*.
        """
        return self.related_parts[rId].slide

    # the rest of this is in tests.parts.test_presentation ------
    def it_provides_access_to_a_related_slide(self, slide_fixture):
        prs_part, rId, slide_ = slide_fixture
        slide = prs_part.related_slide(rId)
        prs_part.related_parts.__getitem__.assert_called_once_with(rId)
        assert slide is slide_

    @pytest.fixture
    def slide_fixture(self, slide_, related_parts_prop_):
        prs_part = PresentationPart(None, None, None, None)
        rId = 'rId42'
        related_parts_ = related_parts_prop_.return_value
        related_parts_.__getitem__.return_value.slide = slide_
        return prs_part, rId, slide_

    @pytest.fixture
    def related_parts_prop_(self, request):
        return property_mock(request, PresentationPart, 'related_parts')
```

eupharis:
Ok cool. Parts make a bit more sense.

The related_parts mock makes sense. I guess `__getitem__` is all we need to do. That's handy. I'll do that from now on.

Sum4196:
Hello @eupharis and @scanny, is there anything I can do to help with this?

scanny:
Hi Michael, we're always happy to have help with things, the main question is whether you have the skills (or, like I was initially, the willingness to develop them quickly :). It's a pretty big library by many people's standards, and fairly complex, so it takes some time to get oriented. Also we maintain pretty high engineering standards in order to keep the library robust at this size with part-time attention.

If none of that scares you away I recommend reading through this full thread and its predecessor #264 and see what you can learn about what it takes. After that you would probably might want to pick a different feature because the coordination overhead would be fairly high and it would be easy to get in each other's way (it's a fairly linear process within a particular feature), but I'll leave that call to @eupharis.

Sum4196:
Your response is much appreciated @scanny, I was reading up on this and it seemed to be more of a linear process as you say which I don't want to interfere with.  I do know that I am not highly skilled at the time being and am learning as I go (I have learned A LOT reading up on this project by the way).  I can possibly take a look at the full thread that you referenced, however, I'm unsure if I would be of much help.  I just figured I would ask anyway :)

dariojr:
Hello! The library is very useful, thanks for your work.
I would like to help in the development of this functionality (header and footer). Although I have experience in the work, but new in collaboration GitHub and high standards.
Do you recommend some documentation to start?
I already read branch 291 and download the last merge to start.
Thank you!

scanny:
@dariojr There is a branch named `feature/header` which contains the committed work so far: https://github.com/python-openxml/python-docx/commits/feature/header

About the last 11 commits on that branch is the header/footer work. If you want to take a look at that and see where you think you want to go next.

Best is probably to get the tests working on your machine as it's all test-driven development.

The unit tests run with py.test and the acceptance tests are run with behave. There is also syntax checking with flake8. Once you get those three installed, the tests are run with:

`make clean & flake 8 & py.test && behave --stop`

from the project directory as the current working directory.

What operating system are you on?

dariojr:
Hi, work in Ubuntu 12.04 tls, I am testing now.

ascendedcrow:
Is there a timeframe for this to go in?

formatically:
Quick question... Why has this not been merged into another release?

khanzf:
python-docx has 41 pull requests and has not updated since Jun 22, 2016.
It might be useful to fork and start a new active branch.

scanny:
@formatically Work stalled on this contributor branch a while back. That often happens on contributed branches. If you want to pick it up by all means do. @khanzf This is a test-driven project, so no commits get merged without the tests. That discourages most casual additions it seems. In my experience only a relative few Python developers are test-driven development (TDD) guys. If you want to add some features by all means open the conversation here. Or if you want to maintain a non-TDD version, fork away :) You might soon be spending your weekends chasing bugs though. That's where indiscriminate committing of pull requests without tests leads in my experience. It might be more productive to become maintainer on this project.

guerda:
@scanny: I'm happy to help to push this feature to release.
What do you see to be done to be able to merge this feature branch? As far as I can see, Travis has been building and testing that branch successfully. Again, I'm happy to help

wheeled:
Hi guys, is this PR working? I’d like to check it out.

Spandan-Madan:
Can this be used to edit the header of a docx file?

If so, can someone please give a quick example for doing so?

Thanks a lot!

scanny:
A client sponsored the headers/footers feature and it is forthcoming in the next release v0.8.8 in the next few days. @ondrej-111 I credited you with several of the commits, although the implementation is somewhat different. You can see the implementation on the `spike-header` branch for comparison if you like. @eupharis I credited you with the analysis and user documentation commits. Thanks to both of you for your contributions to this feature and I'm glad we're finally getting it added to the build :)

eupharis:
@scanny Great news! Thanks for finishing this. Excited to try it out :)

PR #312, opened 2016-08-01
feature: Cell.vAlign property

wich:
- Document analysis and API proposal for Cell.vAlign
- Preliminary acceptance tests for Cell.vAlign


wich:
@scanny please have a look at this one as well and let me know your feedback.

I basically have the implementation available already. How small do you want the commits for this? i.e. should separate commits be made for docx.table._Cell.vAlign, docx.oxml.table.CT_TcPr.vAlign, docx.oxml.table.CT_Tc.vAlign and docx.oxml.table.CT_VerticalJc? Or should I just combine them all in one commit?

docs/dev/analysis/features/table/table-cell.rst
```diff
@@ -7,6 +7,20 @@ properties affecting its size, appearance, and how the content it contains is
...
+Candidate protocol::
+
+    >>> from docx.enum.table import WD_VERTICAL_ALIGNMENT
+    >>> row = table.add_row()
+    >>> row.cells[0]
+    <docx.table._Cell object at 0x...>
+    >>> row.cells[0].vAlign
+    None
+    >>> row.cells[0].vAlign = WD_VERTICAL_ALIGNMENT.CENTER
+
```
scanny: Let's add the alias `WD_ALIGN_VERTICAL` to the `WD_VERTICAL_ALIGNMENT` enumeration 
so folks can write e.g. `WD_ALIGN_VERTICAL.TOP` to correspond to `wdAlignVerticalTop`.
The capitalized snake-case version of the MS API name, as you have here, usually corresponds to the "prefix" of the member names. But in this case, the members all start with `wdAlignVertical..`.

There's a decorator defined for that, used like it is here:
https://github.com/python-openxml/python-docx/blob/master/docx/enum/dml.py#L46

Then you can use the WD_ALIGN_VERTICAL alias here in the import just as though it were a separate enumeration.

----

The property name should be `.vertical_alignment` rather than `.vAlign`. The library convention is to use snake case (no camel case) adapted from the MS API property name (`VerticalAlignment` in this case). We occasionally deviate from that but only when there's a compelling reason.

docs/dev/analysis/features/table/table-cell.rst
```diff
@@ -29,6 +43,15 @@ MS API - Partial Summary
...
 
+WD_VERTICAL_ALIGNMENT Enumeration
+---------------------------------
+
+* wdAlignVerticalTop (0) Top vertical alignment.
+* wdAlignVerticalCenter (1) Center vertical alignment.
+* wdAlignVerticalJustify (2) Justified vertical alignment.
+* wdAlignVerticalBottom (3) Bottom vertical alignment.
+
+
```
scanny: We should add an example using vertical alignment to the XML specimen below. We'll want a known-correct example handy when we get into the implementation :)

docx/enum/table.py
```diff
@@ -69,3 +69,35 @@ class WD_TABLE_DIRECTION(Enumeration):
             'in the rightmost position.'
         ),
     )
+
+
+class WD_VERTICAL_ALIGNMENT(XmlEnumeration):
```
scanny: It looks like this enumeration can also be used by sections and perhaps shapes. So we should create a new module named 'shared.py' in this same directory and place this enumeration there.

----

We'll need the documentation page for this as well. I got tired of writing those by hand so you'll find a property on the enumeration named `.__docs_rst__` that generates the exact text of the enumeration documentation page. It goes under /docs/api/enum I believe and the filename is the same as the __ms_name__ below. You'll need to add a line for the new page to the index.rst page in that same directory.

```python
>>> from docx.enum.shared import WD_ALIGN_VERTICAL as wd
>>> print(wd.__docs_rst__)
# then copy and paste or whatever :)
```

features/steps/table.py
```diff
@@ -351,6 +366,15 @@ def then_the_reported_column_width_is_width_emu(context, width_emu):
...
 
+@then('the reported vertical alignment of the cell is {vAlign}')
+def then_the_reported_vertical_alignment_of_the_cell_is_vAlign(context, vAlign):
+    expected_vAlign = {'None': None, 'bottom': WD_VERTICAL_ALIGNMENT.BOTTOM}[vAlign]
+    actual_vAlign = context.cell.vAlign
+    assert actual_vAlign == expected_vAlign, (
+        'expected %s, got %s' % (expected_vAlign, actual_vAlign)
+    )
+
+
```
scanny: Generally in these roles I just use the word "value" (rather than 'vAlign' in this case). There are a number of reasons, but one is that it can be consistent across different tests so makes it easier to read and make sense of quickly. Also, like in this case, if you have to change the name for some reason these don't have to change :)

features/tbl-cell-props.feature
```diff
@@ -4,6 +4,29 @@ Feature: Get and set table cell properties
...
+  @wip
+  Scenario Outline: Get cell vertical alignment
+    Given a table cell having a vertical alignment of <vAlign-setting>
+     Then the reported vertical alignment of the cell is <reported-vAlign>
+
+    Examples: Table cell vertical alignment settings
+      | vAlign-setting        | reported-vAlign |
+      | no explicit setting   | None            |
+      | bottom                | bottom          |
+
```
scanny: On these I've arrived at the convention of using the member name (without its enumeration "prefix" instead of a descriptive name. So "BOTTOM" in this case instead of "bottom". It has the added benefit that once you've ruled out one of the exceptional terms like "None" or "no explicit setting" the rest can be resolved in the steps code using e.g. `getattr(WD_ALIGN_VERTICAL, member)`, so you don't have to spell everything out in a dict.

scanny:
Ok @wich, I've left comments for you on this one. This looks really good, I'm sure we can merge these once you made updates and we can move on to the implementation :)

docs/dev/analysis/features/table/table-cell.rst
wich: I'll just rename the enum to WD_ALIGN_VERTICAL, I have no preference of one over the other.

wich:
@scanny your comments have been addressed, if you could have a look.

docs/dev/analysis/features/table/table-cell.rst
```diff
@@ -29,6 +43,15 @@ MS API - Partial Summary
 * WordWrap
 
 
+WD_ALIGN_VERTICAL Enumeration
+---------------------------------
+
+* wdAlignVerticalTop (0) Top vertical alignment.
+* wdAlignVerticalCenter (1) Center vertical alignment.
+* wdAlignVerticalJustify (2) Justified vertical alignment.
```
scanny: @wich I'm not finding this Justify option in the MS-defined enumeration, although I do see that 'both' is a defined option for the `w:val` attribute value. Did you find the wdAlignVerticalJustify member defined somewhere or is this just a best guess so far of what that indicates as far as behavior in Word?

Also, have you tried out what a table cell having the 'both' value does? I'm supposing it just ignores that particular value.

scanny:
Okay, these two commits are merged on feature/cell-vert-align:
https://github.com/python-openxml/python-docx/commits/feature/cell-vert-align

Note the enumeration is WD_CELL_ALIGN_VERTICAL (corresponding to WdCellAlignVertical) for table cells. I made the necessary changes.

You should be good to go on the implementation.

PR #326, opened 2016-10-23
Enable python-docx to add border

xxxspy:
Enable python-docx to add border with `docx.Document('test table.docx').tables[0].rows[2].add_border('left',color='e1a3a7')`


akonsta:
Is there a reason this has not been merged?  I am very interested in this feature.

PR #329, opened 2016-10-26
east asia, chinese font family

leonzhu211:
My test code is like this:
    styles = document.styles
    s1 = styles.add_style('s1', WD_STYLE_TYPE.PARAGRAPH)
    s1.font.size = Pt(40)
    s1.font.name = u'楷体_GB2312'
    document.add_paragraph(u'中文内容', style = s1)

The chinese font family name has no effect.
So I create a docx file with the goal content, unzip it and compare the xml with that generated by python-docx. I find that the difference is the following fragment:

<w:rFonts w:eastAsia="楷体_GB2312" w:asciiTheme="minorAscii" w:hAnsiTheme="minorAscii"/></w:rPr>

That's why I add the property 'eastAsia'.

### DIFF
--- a/docx/oxml/text/font.py
+++ b/docx/oxml/text/font.py
@@ -32,6 +32,7 @@ class CT_Fonts(BaseOxmlElement):
     """
     ascii = OptionalAttribute('w:ascii', ST_String)
     hAnsi = OptionalAttribute('w:hAnsi', ST_String)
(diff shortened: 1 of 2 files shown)


fengdasuk19:
The unit test code has been updated in PR https://github.com/python-openxml/python-docx/pull/781 .

Issue #167, opened 2015-03-29
Create newsletter columns within document


mike-charles:
Is it possible to create columns in a document? I don't mean table columns, I mean columns within the body text, according to [this](https://support.office.com/en-sg/article/Create-newsletter-columns-3a00e688-253d-456e-9481-661fdd5b724b#bm3) documentation.

For example:
![](http://www.trickyways.com/wp-content/uploads/2010/05/3-columns-in-word-2.png)

scanny:
There is not yet API-level support for this in python-docx. Was that what you were asking?

mike-charles:
Yes that's what I was wondering. Thanks! Should I leave this open for the future?

scanny:
Yes, please

gimbo:
+1 for this feature

electron378:
Is anyone working on this feature now or is there any show-stopper in development? I had an urgent need in this feature and it took less than 20 minutes to figure out a working solution so I'd share if it helps anyone. Is there anything fundamentally wrong with the solution below and if not - what would be the right way to integrate it (apart from the tests)? 

For those needing this now here is a quick solution:
```python
WNS_COLS_NUM = "{http://schemas.openxmlformats.org/wordprocessingml/2006/main}num"


def set_number_of_columns(section, cols):
    """ sets number of columns through xpath. """
    section._sectPr.xpath("./w:cols")[0].set(WNS_COLS_NUM, str(cols))
```
use example:
```python
from docx import Document
from docx.enum.section import WD_ORIENT, WD_SECTION


doc = Document()
doc.add_paragraph("this page should be normal 1 column")
doc.add_page_break()
section = doc.add_section(WD_SECTION.NEW_PAGE)
set_number_of_columns(section, 2)
doc.add_paragraph("this page should have 2 columns and it seems to work ok")
```

scanny:
Well, just offhand, I'd say you'd need to account for the case when `w:cols` is not already present, if it's an optional element, which, of course, most of them are. To determine that, you'd need to have the relevant parts of the XML schema, which is part of what goes on the enhancement proposal, aka. "analysis page".

Then the element name is more elegantly handled by the internal function `qname()`, like `qname('w:num')`. That function lives in `docx.oxml` I believe.

Also, of course, it would need to become a method on section, or more likely a set of properties or perhaps an object of its own, depending on the scope of things that can eventually be done to columns, like perhaps set their width, margin, or space between.

[...]

PR #355, opened 2017-01-29
feature analysis: section columns #167

electron378:
Provides analysis page for section-columns feature:
- XML schema analysis
- candidate API
- specimen XML

### DIFF
--- a/docs/dev/analysis/features/section-columns.rst
+++ b/docs/dev/analysis/features/section-columns.rst
@@ -1,0 +1,171 @@
+Section with multiple Columns
+=============================
+
+Word provides means for the end-user to create document sections with multiple
+columns. By default, all columns within a section are of the same width and
+equally spaced. A user may customize width and spacing (margin-right) of all or
+any individual column. Also a separator line may be displayed.
+
+Schema analysis
+---------------
+
+In a WordprocessingML document, the number of section columns and their geometry
+is defined within `w:cols` element (optional, typed as `CT_Columns`) of a
+`sectPr` structure. The type spec follows:
+
+.. highlight:: xml
+
+::
+
+  <xsd:complexType name="CT_Columns">
+    <xsd:sequence minOccurs="0">
+      <xsd:element name="col" type="CT_Column" maxOccurs="45"/>
+    </xsd:sequence>
+      <xsd:attribute name="equalWidth" type="s:ST_OnOff"/>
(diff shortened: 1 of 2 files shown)


electron378:
@scanny, could you please review the proposed API? It would help a lot as I may get some time to move this feature forward over the weekend. Thanks!

scanny:
Hi @electron378, thanks for this, I've had a look over it.

I'd say it's a great start. The main thing is it doesn't look like you've referred to the Microsoft API for Word (MS API). That turns out to be a critical piece of the puzzle for avoiding mis-steps in defining the python-docx API.

If it's not immediately apparent how to do what I'm defining using the MS API, I search on something like "vba word section columns" and that will generally get me close enough to find what I'm looking for in the Word API reference: http://msdn.microsoft.com/en-us/library/office/ff837519.aspx

In this case, it turns out to be critical, because it points to a shared object, a `PageSetup` object I believe, that appears both in sections and perhaps for the document overall. So the protocol would look something like:

```python
section = document.sections[-1]
page_setup = section.page_setup
text_columns = page_setup.text_columns
text_column = text_columns[0]
text_column.width = Inches(2)
text_column.space_after = Inches(0.25)
```

This is the neighborhood in which to look in the Word API:

https://msdn.microsoft.com/en-us/library/office/ff194295.aspx
https://msdn.microsoft.com/en-us/library/office/ff191765.aspx
https://msdn.microsoft.com/en-us/library/office/ff840256.aspx
etc. ...

It looks like there is also a `Document.PageSettings` property in the MS API, so if we add that later it can reuse the PageSettings objects we create here.

In general, I've found it best to mirror the MS API whenever possible, using pythonicized names. There are two parts to the rationale. First is that the designers put a lot of thought into it, more than we're likely to, and what they have works as a complete whole. This is a big plus, because we're adding on individual features incrementally, and its easy to "re"-design yourself into a box where you later have to change the API. Changing the API is the one big no-no in a library. Folks build things that depend on the API and if you change it you break their code.

The second reason is because at least some folks are already familiar with the MS API, so might as well not make it any harder for them than necessary.

scanny:
Also make sure you include the relevant XML Schema excerpts. The XML schemas are in the code tree here:
https://github.com/python-openxml/python-docx/tree/master/ref/xsd

Most of the elements for this will probably be in `wml.xsd`. Check out the other analysis pages for examples of proper layout.

PR #376, opened 2017-03-15
docs: document pictures_in_runs analysis

DKWoods:
Analysis for making image data available in Runs when reading docx files.

### DIFF
--- a/docs/dev/analysis/features/text/pictures-in-runs.rst
+++ b/docs/dev/analysis/features/text/pictures-in-runs.rst
@@ -1,0 +1,105 @@
+Shapes in Runs
+==============
+
+This document is aimed at accessing image information when reading docx files.  Adding image information as part of creating docx files is handled elsewhere.
+
+A Run <w:r> can contain a Drawing <w:drawing>, which can can contain an Inline <wp:inline>, which can contain a Graphic <a:graphic>, which can contain graphicData <a:graphicData>, which can contain a Picture <pic:pic>.  A Picture can contain a blipFill <pic:blipFill> and a "Shape Properties" spPr <pic:spPr>.  A blipFill contains a Blip <a:blip> with an embed property that provides the graphic's Resource ID (rId) needed to look the image up in document.part.rels.  The spPr contains informtion about the image's size.  All of this except the Drawing element are documented in docs/dev/analysis/features/shapes/shapes-inline.rst.
+
+
+Word Behavior
+-------------
+
+In theory, a Run can contain multiple Drawings.  In practice, I have not found a way to create a Run with more than a single Drawing.
+
+
+Protocol
+--------
+
+.. highlight:: python
+
+To stay parallel with document.add_picture() behavior, which creates an Inline object, the Run should have an Inlines property, producing a list of Inline objects.
+
+  >>> inlines = run.inline_lst
+  >>> inline = inlines[0]
+  >>> inline
+  <CT_Inline '<wp:inline>' at 0x43de510>
+  
+From here, we can access properties which characterize the image.
+
+  >>> rId = inline.graphic.graphicData.pic.blipFill.blip.embed
+  >>> rId
+  'rId7'
+  >>> cx, cy = inline.graphic.graphicData.pic.spPr.cx, inline.graphic.graphicData.pic.spPr.cy
+  >>> cx.inches, cy.inches
+  (4.61111111111, 2.33333333333)
+  >>> document.part.rels[rId]._target.image.content_type
+  'image/png'
+  >>> rels[rId]._target.image.filename, rels[rId]._target.image.ext
+  ('image.png', 'png')
+  >>> originalx, originaly = rels[rId]._target.image.width, rels[rId]._target.image.height
+  >>> originalx, originaly
+  (4.61111111111, 2.33333333333)
+  >>> image_data = rels[rId]._target.image.blob
+  >>> len(image_data)
+  23442
+  
+This interface, while (mostly) following Microsoft's object model (except for leaving out the Drawing level), seems unfriendly to the python_docx user.  Therefore, I would recommend the following additional interface elements.
+
+  >>> pictures = run.picture_lst
+  >>> picture = pictures[0]
+  >>> picture.rId
+  'rId7'
+  >>> picture.cx.inches, picture.cy.inches, picture.originalx.inches, picture.originaly.inches
+  (4.61111111111, 2.33333333333, 4.61111111111, 2.33333333333)
+  >>> picture.content_type
+  'image/png'
+  >>> picture.filename, picture.ext
+  ('media/image.png', 'png')
+  >>> image_data = picture.image_data
+  >>> f = open('image_export.%s' % picture.ext, 'wb')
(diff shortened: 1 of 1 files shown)


docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
+Shapes in Runs
```
scanny: There is an existing set of pages for inline shapes: http://python-docx.readthedocs.io/en/latest/dev/analysis/features/shapes/shapes-inline.html (this link is the middle one of three pages)

This should just be an elaboration of the one I linked to just above. Most of the bits like schema excerpts and so on are the same and don't need to be repeated.

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+==============
+
+This document is aimed at accessing image information when reading docx files.  Adding image information as part of creating docx files is handled elsewhere.
+
+A Run <w:r> can contain a Drawing <w:drawing>, which can can contain an Inline <wp:inline>, which can contain a Graphic <a:graphic>, which can contain graphicData <a:graphicData>, which can contain a Picture <pic:pic>.  A Picture can contain a blipFill <pic:blipFill> and a "Shape Properties" spPr <pic:spPr>.  A blipFill contains a Blip <a:blip> with an embed property that provides the graphic's Resource ID (rId) needed to look the image up in document.part.rels.  The spPr contains informtion about the image's size.  All of this except the Drawing element are documented in docs/dev/analysis/features/shapes/shapes-inline.rst.
```
scanny: This paragraph just reiterates in prose what the schema states unambiguously. If you want this for your technical notes, that's fine, but no need for it here.

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+Word Behavior
+-------------
+
+In theory, a Run can contain multiple Drawings.  In practice, I have not found a way to create a Run with more than a single Drawing.
```
scanny: This is a useful observation. We should add a Word behavior section to the existing Inline Shapes page I linked to and put this there.

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+
+
+Protocol
+--------
+
+.. highlight:: python
+
+To stay parallel with document.add_picture() behavior, which creates an Inline object, the Run should have an Inlines property, producing a list of Inline objects.
```
scanny: Be careful about mixing API object names and XML element names. The collection type here is `InlineShapes` and should already exist because its available from `Document` on `Document.inline_shapes`. I think you'll find the MS API property is `Document.InlineShapes`: https://msdn.microsoft.com/en-us/library/office/ff844996.aspx

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+
+  >>> inlines = run.inline_lst
+  >>> inline = inlines[0]
+  >>> inline
+  <CT_Inline '<wp:inline>' at 0x43de510>
```
scanny: The protocol expresses the behavior of the library from the perspective of the user (a developer). The user doesn't deal directly with the oxml layer, that's an internal implementation detail. So minimally, this sort of thing belongs in a different section. I don't typically document the oxml design, I just work it out when I get there, mostly because it's so convention-driven there's usually not a lot to decide on. But if you want this, just segregate it into a section lower down, maybe 'Oxml Implementation' or something.

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+  >>> document.part.rels[rId]._target.image.content_type
+  'image/png'
+  >>> rels[rId]._target.image.filename, rels[rId]._target.image.ext
+  ('image.png', 'png')
+  >>> originalx, originaly = rels[rId]._target.image.width, rels[rId]._target.image.height
+  >>> originalx, originaly
+  (4.61111111111, 2.33333333333)
+  >>> image_data = rels[rId]._target.image.blob
+  >>> len(image_data)
+  23442
```
scanny: Look and see what API InlineShape already has. You'll be building on top of that, if you need to at all. Much of this looks like it would be encapsulated by the oxml layer and not be exposed by the API.

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+  (4.61111111111, 2.33333333333, 4.61111111111, 2.33333333333)
+  >>> picture.content_type
+  'image/png'
+  >>> picture.filename, picture.ext
+  ('media/image.png', 'png')
+  >>> image_data = picture.image_data
+  >>> f = open('image_export.%s' % picture.ext, 'wb')
+  >>> f.write(image_data)
+  >>> f.close()
+  (produces a PNG image file named 'image_export.png' containing the image)
```
scanny: Let's get the base API worked out before we go about extending it with our own creations. I've found freelancing the API to be a very risky business as the MS API has a strong tendency to have good reasons why it's the way it is and those usually show up after you've written a lot of code that then needs to get scrapped :) You can leave this in, maybe as a "tentative" section, and we can take a closer review once you've got the rest of it fairly clear.

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+          <pic:spPr>
+            (information about image offset and size)
+          </pic:spPr>
+        </pic:pic>
+      </a:graphicData>
+    </a:graphic>
+  </wp:inline>
+</w:drawing>
+
+See documentation of CT_Inline, CT_GraphicalObject, CT_GraphicalObjectData, CT_Picture, CT_BlipFillProperties, and CT_Blip for more information.
```
scanny: I know that some of the other analysis pages have this 'Minimal XML' section, but it's a practice I've moved away from. It turns out, again, that Word usually has a good reason for inserting the elements it does, and those reasons often only show up later, making it more painful to add them in.

Instead of our best-guess at minimal XML, we should put in a snippet of the XML that Word actually inserts. So create a paragraph, insert a picture and save. Then snippet that `w:p` here so we can see not only the drawing, but the context of the run that contains it (and the paragraph, just for good measure; might make sense later to add a `Paragraph.InlineShapes` property, never know :)

docs/dev/analysis/features/text/pictures-in-runs.rst
```diff
@@ -0,0 +1,105 @@
...
+                   
+                   
+Resources
+---------
```
scanny: The XML Schema excerpt is an absolutely mandatory section in an analysis document. You can't get the oxml right without it. The Inline shape page this content will be added to already has some excerpts. You'll want to make sure it has everything you're going to need.

Issue #85, opened 2014-08-10
feature: Paragraph.text includes hyperlink text


SebasSBM:
When getting Document.paragraphs objects, their text method doesn't include hyperlinks in the output. The one with this problem posted his question here:

http://stackoverflow.com/questions/25228106/how-to-extract-text-from-an-existing-docx-file-using-python-docx/25228787#25228787

I've been reading the documentation of python-docx for several hours and didn't find any property or method useful to resolve this. Maybe some class should be created, or some methods should be appended to an existing class to achieve this.

I barely know something about python-docx API. I knew of it's existence trying to help some people in stackoverflow.com with their problems. I don't even know how Windows' DOCX format works (I tried to open it with an hexadecimal editor to try to figure it out, and I just don't get it :P). But I'm skilled with logic problems and I've got good skills with Python scripting. I'd like to help if there's something I can do.

mikkleini:
Hi,
python-docx is a great tool that i just discovered. However, i also faced the missing hyperlink text problem right away.
Please solve it.

mikkleini:
Ok, i think i faced the Python 2/3 issue. Here's something which works on Python 3.4:

```python
def remove_hyperlink_tags(xml):
    import re
    text = xml.decode('utf-8')
    text = text.replace("</w:hyperlink>","")
    text = re.sub('<w:hyperlink[^>]*>', "", text)
    return text.encode('utf-8')
```

funkycode:
any schedule for that one?

herrera78:
I used the code at https://gist.github.com/etienned/7539105#file-extractdocx-py and supplied the path and get hyperlink texts as I need

desecho:
@scanny Please add your comments concerning #377. It seems to work correctly and provides the functionality described in this issue.

sanjuroj:
I'd like to take a crack at adding this feature, but I'm a relatively inexperienced developer, especially on these kinds of large projects, and I'm going to have a lot of questions along the way.  If it seems like it would be too much to help me in this effort, please let me know and I'll leave it be.  

[...]

PR #377, opened 2017-03-17
Feature: Paragraph.text includes hyperlink text

desecho:
Fixes issue https://github.com/python-openxml/python-docx/issues/85

### DIFF
--- a/docx/oxml/__init__.py
+++ b/docx/oxml/__init__.py
@@ -7,6 +7,8 @@ corresponding to Open XML elements.
 
 from __future__ import absolute_import
 
+import re
+import six
 from lxml import etree
 
 from .ns import NamespacePrefixedTag, nsmap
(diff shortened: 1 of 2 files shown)


scanny:
@desecho I'm not sure what you're trying to do with this one bit I won't merge it. Probably better to discuss before you go any further.

scanny:
Ok, I see what you're trying to do now.

A few serious problems:

1. There's no documented analysis of the situation and the proposed fix. Changes are initiated by proposing an enhancement (using an enhancement proposal, aka. analysis page). There's a lot to discuss and resolve before getting to work, foremost of which is the API additions/changes that will be made since those are very difficult to change once set.

2. There are not acceptance or unit tests. All successful merges include tests.

3. You're making destructive changes at the highest possible level. You shouldn't make destructive changes (unless asked for by the user) and you should make changes as locally as possible. Removing the `w:hyperlink` element changes the document by essentially removing all hyperlinks (and the other information they contain). It's even worse than this (relationship links are left broken, etc.), but this is enough. This makes it a non-starter.

4. You're doing this by modifying `parse_xml()`, which is a fundamental function used throughout the library. This incurs risk in every module that uses it, which is many. Wrong use of a broadly used function.

5. You've added an unwelcome dependency. We don't use `six` on this project. See the `compat` module to see how we take care of those things.

So perhaps you understand my reply better now. This is the wrong way to fix the issue, in a lot of ways.

Issue #276, opened 2016-03-14
how to read paragraphs AND tables?


bit111:
Hi,
I have read the entire discussion on [issue #40](https://github.com/python-openxml/python-docx/issues/40) but the solutions do not work with 0.8.5 release (I think because are dated solutions).
This is my problem: I have a large docx to read with more than 400 pages. In this document I have some data in rows and some data in tables.
How I can do to read paragraphs and tables in the order they appears in the doc?
Thanks

scanny:
What code are you using for iter_block_items() and what error are you getting?

bit111:
I am using this code
```
def main():
	doc = Document('file1.docx')
	#output = open("output_"+cur_date+"_.txt", "w")
	parent_elem = doc.element
	for block in iter_block_items(doc) :
		print(block.text)


def iter_block_items(parent):
    """
    Yield each paragraph and table child within *parent*, in document order.
    Each returned value is an instance of either Table or Paragraph. *parent*
    would most commonly be a reference to a main Document object, but
    also works for a _Cell object, which itself can contain paragraphs and tables.
    """
    if isinstance(parent, Document):
        parent_elm = parent.element
    elif isinstance(parent, _Cell):
        parent_elm = parent._tc
    else:
        raise ValueError("something's not right")
		
    parent_elm = parent.element

    for child in parent_elm.iterchildren():
        if isinstance(child, CT_P):
            yield Paragraph(child, parent)
        elif isinstance(child, CT_Tbl):
            yield Table(child, parent)

```
And the error is:

> Traceback (most recent call last):
>   File "beta-iterblock.py", line 47, in <module>
>     main()
>   File "beta-iterblock.py", line 21, in main
>     for block in iter_block_items(doc) :
>   File "beta-iterblock.py", line 32, in iter_block_items
>     if isinstance(parent, Document):
> TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types

scanny:
Why do you have the Document import commented out? What does the 'Document' name resolve to?

bit111:
Because if I use these lines of import
```
from docx.document import Document
from docx.oxml.table import CT_Tbl
from docx.oxml.text.paragraph import CT_P
from docx.table import _Cell, Table
from docx.text.paragraph import Paragraph
```
I have this error:
```
Traceback (most recent call last):
  File "beta-iterblock.py", line 43, in <module>
    main()
  File "beta-iterblock.py", line 15, in main
[...]

PR #395, opened 2017-05-03
document.story property, contains Paragraph and Table in document order

belegnar:
#276 
a dirty solution, but it works fine

### DIFF
--- a/docx/blkcntnr.py
+++ b/docx/blkcntnr.py
@@ -66,6 +66,23 @@ class BlockItemContainer(Parented):
         from .table import Table
         return [Table(tbl, self) for tbl in self._element.tbl_lst]
 
+    @property
+    def story(self):
+        """
+        A list containing paragraphs and tables in document order
+        """
+        from .table import Table
+        from docx.oxml.text.paragraph import CT_P
+        from docx.oxml.table import CT_Tbl
+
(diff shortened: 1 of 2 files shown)


docx/blkcntnr.py
```diff
@@ -66,6 +66,23 @@ def tables(self):
...
+        """
+        A list containing paragraphs and tables in document order
+        """
+        from .table import Table
+        from docx.oxml.text.paragraph import CT_P
+        from docx.oxml.table import CT_Tbl
+
+        def make_element(el, s):
+            if isinstance(el, CT_P):
+                return Paragraph(el, self)
```
flutefreak7: Did you mean for this to refer to "self" where line 82 refers to "s"

timuram: Does it matter? `make_element` is called once with `(.., self)` and no other usage to this function is possible. So it may be either `s` or `self`

belegnar: Does it matter? `make_element` is called once with `(.., self)` and no other usage to this function is possible. So it may be either `s` or `self`

jfthuong: @belegnar  Why putting a  `s` argument then? Could you not just use the `self` from the call to `story` method?

PR #429, opened 2017-08-31
Add method to add picture to table cell. See #1.

eclacfo:
Unless I'm mis-reading the documentation there is not currently support for this.  Having read the _Cell class code, I noticed the add_picture method was missing.  I've added the method for my own ease of use, perhaps it may be useful to someone else as well.


martinmde:
I re-invented that method, too,  works  nicely.
Just insert a run  which allows  to use  add_picture

PR #445, opened 2017-11-11
Feature/bookmarks

Benjamin-T:
Dear @scanny,

I've created a pull request like you requested. I'm trying to figure out how to connect this pull request with the issue I created: #425 

regards,

- Ben

PR #539, opened 2018-08-25
Feature/bookmarks

scanny:
This PR provides a place for me to comment on differences between merged commits and those submitted on PR #445.


features/doc-access-collections.feature
```diff
@@ -27,3 +27,9 @@ Feature: Access document collections
   Scenario: Access the tables collection of a document
     Given a document having three tables
      Then document.tables is a list containing three tables
+
+
+  @wip
+  Scenario: Document.bookmarks
+    Given a Document object as document
+     Then document.bookmarks is a Bookmarks object
```
scanny: The project's style for acceptance tests has evolved somewhat since many of these were written. This one reflects the current style, which adopts the perspective that the "user" in these user acceptance tests is a _developer_ using the _API_. So the scenario name is usually terse like this, often {object}.{method or property}, sometimes with "getter" or "setter" appended when it's two aspects of a property.

The given statement in this style mentions the "variable" name used for the object in an "as" clause at the end.

This test is quite simplified from your original version. This reflects that there are three "levels" or "steps" that are tested, each individually:

1. A document object has a `.bookmarks` property that returns (unconditionally and idempotently in this case) a `Bookmarks` object.

2. A Bookmarks object has certain behaviors (len, indexed access, interable, etc.)

3. An individual _Bookmark_ object has certain behaviors.

These are each tested separately and not mixed in one test.

features/steps/document.py
```diff
@@ -27,6 +27,11 @@ def given_a_blank_document(context):
     context.document = Document(test_docx('doc-word-default-blank'))
 
 
+@given('a Document object as document')
+def given_a_Document_object_as_document(context):
+    context.document = Document(test_docx('doc-default'))
```
scanny: No need for a special document file yet, and better that we don't introduce one for this because bookmarks should work even when the document has none. We'll add your test file in a later commit.

features/steps/document.py
```diff
@@ -162,6 +167,14 @@ def when_I_change_the_new_section_layout_to_landscape(context):
 
 # then ====================================================
 
+@then('document.bookmarks is a Bookmarks object')
+def then_document_bookmarks_is_a_list_containing_three_bookmarks(context):
+    bookmarks = context.document.bookmarks
+    actual = bookmarks.__class__.__name__
+    expected = 'Bookmarks'
+    assert actual == expected, 'bookmarks is a %s object' % actual
```
scanny: While it may seem a trivial test, all we want to do here is establish we got a Bookmarks object on this call. Any particular behaviors of that bookmarks object is for another test. These "..is an X object" tests are quite effective at driving (and defending) the top-level behaviors they're designed to test.

docx/bookmark.py
```diff
@@ -0,0 +1,11 @@
...
+
+"""Objects related to bookmarks."""
+
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+
+class Bookmarks(object):
+    """Sequence of |Bookmark| objects."""
```
scanny: Each commit should have a test (occasionally two) and the code to make it pass. And, all tests should pass on every commit (although @wip acceptance tests are skipped until they pass and don't count in the "all-tests" condition). Only the code required to pass the test is included in the commit.

Accordingly, there is no separate commit to add this module. It's added in its minimum form when first required by a test. It was needed in this commit so it could be mocked (or whatever) so it appears here. Note that it has no dependencies, in particular it inherits from `object` here; subclassing `Sequence` will be added when we need its sequence behaviors for a subsequent test.

docx/document.py
```diff
@@ -8,11 +8,11 @@
...
-from .blkcntnr import BlockItemContainer
-from .enum.section import WD_SECTION
-from .enum.text import WD_BREAK
-from .section import Section, Sections
-from .shared import ElementProxy, Emu
+from docx.blkcntnr import BlockItemContainer
+from docx.enum.section import WD_SECTION
+from docx.enum.text import WD_BREAK
+from docx.section import Section, Sections
+from docx.shared import ElementProxy, Emu, lazyproperty
```
scanny: In the past I used relative imports (e.g. '.' and '..' prefixes) in this project, but have since determined this is not best practice. So whenever I have to add an import I fix the others in that module. Note they are sorted in alphabetical order at the top of the module and items within the line as well. This makes the position for a new import predictable and easy to review.

docx/document.py
```diff
@@ -100,6 +100,17 @@ def add_table(self, rows, cols, style=None):
...
+    @property
+    def bookmarks(self):
+        """|Bookmarks| object providing access to |Bookmark| objects.
+
+        This feature is PARTIAL and EXPERIMENTAL. The current implementation
+        only provides access to bookmarks in the main document story.
+        Conflicts may arise if a bookmark is added with the same name as one
+        appearing in another story such as a header or footer etc.
+        """
+        return self._body.bookmarks
```
scanny: Because of the multi-story scope of bookmarks, I'm inclined to delegate ownership of the actual Bookmarks object to _Body, like `.paragraphs` does with those. Eventually this `Document`-level bookmarks may become an "aggregate" object that contains bookmarks from all stories in the document; so this seam will allow us to add that later (and incrementally) with a minimum of disruption.

docx/document.py
```diff
@@ -205,6 +216,11 @@ def __init__(self, body_elm, parent):
         super(_Body, self).__init__(body_elm, parent)
         self._body = body_elm
 
+    @lazyproperty
+    def bookmarks(self):
+        """|Bookmarks| object providing access to main story bookmarks."""
+        raise NotImplementedError
```
scanny: When needed for mocking purposes, a new method/property is added as a "stub". It nevertheless always has a complete docstring (that is formatted according to PEP 257). Its only line is `raise NotImplementedError` which causes the acceptance test to fail at just the right spot when it's next in line for implementation.

tests/test_document.py
```diff
@@ -88,6 +89,16 @@ def it_can_save_the_document_to_a_file(self, save_fixture):
...
 
+    def it_provides_access_to_its_bookmarks(
+            self, body_prop_, body_, bookmarks_):
+        document = Document(None, None)
+        body_prop_.return_value = body_
+        body_.bookmarks = bookmarks_
+
+        bookmarks = document.bookmarks
+
+        assert bookmarks is bookmarks_
```
scanny: In the past, I used to _always_ have a fixture for each test. Now I only use a test fixture when there are multiple cases (i.e. fixture needs `params=[]`). It seems like it makes it easier to have all the fixture-making right there with the test when that's possible. This doesn't apply to fixture components though (single mocks signified by a trailing underscore ('_') in their name. Those are just used directly by the test rather that a fixture.

tests/test_document.py
```diff
@@ -291,12 +306,12 @@ def body_(self, request):
         return instance_mock(request, _Body)
 
     @pytest.fixture
-    def _block_width_prop_(self, request):
-        return property_mock(request, Document, '_block_width')
```
scanny: I maintain fixture components in strict alphabetical order by name. This one was out of order, which explains this change not directly related to this test. Keeping them alphabetical speeds finding the one you're after when scanning the source, especially when code-folding and search aren't handy.

scanny (commented):
@Benjamin-T I've left some explanatory comments where the merged commits differ from those submitted with your PR.

Benjamin-T:
@scanny #445 has the latest master as base, shouldn't we rebase this as well and see if we can get some new commits merged into this branch? 
regards, 

Ben

Benjamin-T:
@scanny
What would it take to get the bookmarks feature added to the library? If you are very time constrained, perhaps in small steps. Or otherwise, contact me directly to see if it can be done via a sponsorship.
Regards,
Ben

scanny:
Hi Benjamin, I did some work on this over the weekend:

* I rebased my `feature/bookmarks` onto the latest master, which contains the work recently completed on adding headers and footers.

* I added about seven commits from your branch, up to `bmk: add Document.start_bookmark()`.

Please review this to understand the changes I made and why, then rebase your branch on my `feature/bookmarks` such that your remaining commits proceed from those that have been committed here (and do not also include commits already on my `feature/bookmarks` branch. You may prefer to do this by cherry-picking; any way that works for you is fine with me.

Then I'll take another look.

scanny:
@Benjamin-T Hi Benjamin, I did some work on this over the weekend:

* I rebased my `feature/bookmarks` onto the latest master, which contains the work recently completed on adding headers and footers.

* I added about seven commits from your branch, up to `bmk: add Document.start_bookmark()`.

Please review this to understand the changes I made and why, then rebase your branch on my `feature/bookmarks` such that your remaining commits proceed from those that have been committed here (and do not also include commits already on my `feature/bookmarks` branch. You may prefer to do this by cherry-picking; any way that works for you is fine with me.

Then I'll take another look.

Benjamin-T:
@scanny 
Thanks for taking the time and adding some of my commits. Dit you also notice that I've updated the feature analysis document? I've tried to answer some of the remaining questions. If you find the answers satisfactory, perhaps that document can be added to this PR as well.

Benjamin-T:
Dear @scanny 

I've rebased my branch and added some commits. I've also updated the branch with an additional acceptance test. If you could take a look and see if they can be merged that would be much appreciated. 

The last commit `bmk: add BaseStoryPart.iter_story_parts()` is a refactor proposal:
If we change the location of the 'iter_story_parts() method from the `DocumentPart` to the `BaseStoryPart` the header and footer parts can access the same method. (as well as any other storypart we might imlement in the future i.e: Comments and Endnotes.)

I haven't changed it in my current branch in order to make it more obvious what I mean.

Curious about your thoughts on the matter.

regards ,

Ben

Benjamin-T:
@scanny 

There are some pending commits in my feature branch. Please let me know when you've looked at them so we can move forward.

thanks for time

- Ben

Benjamin-T:
Hi Steve, 

I somehow missed your new commits to this PR - Great stuff!

I've experimented a bit with it and discovered that even though the bookmarks added do show up in the word editor, they are somewhat different. 

The main issue is that python-docx places them at the same level as the paragraph elements, whereas the word editor inserts the bookmark elements on a paragraph level. I've checked this for the body and header stories and assume that this holds for the other story-parts as well. 

The current implementation results in this xml:
```python
from docx import Document
doc = Document()
bmk = doc.start_bookmark('test')
doc.end_bookmark(bmk)
doc.save('test_bookmark.docx')
```
```xml
<w:document>
  <w:body>
    <w:bookmarkStart w:name="test" w:id="1" />
    <w:bookmarkEnd w:id="1" />
  </w:body>
</w:document>
```
Whereas the word editor creates this xml:
(Create document using python-docx --> click insert bookmark --> save file)

```xml
<w:document>
  <w:body>
    <w:p>
      <w:bookmarkStart w:name="test" w:id="0" />
      <w:bookmarkEnd w:id="0" />
    </w:p>
  </w:body>
</w:document>
```

scanny:
@Benjamin-T I think the safest bet is to determine what the Microsoft API for Word does and reproduce that behavior. Because details like this are not documented (publicly at least), the only good assurance I've found is to duplicate the MS API behavior, which is pretty good assurance in my experience, although sometimes a mystery why it needs to be the way it is.

This is an example of why the analysis document is so important, in particular doing the thorough analysis work needed to properly create that document. I of course learned that the hard way, by reimplementing several features early on, but this is definitely a "teaching moment" opportunity :)

There are a couple ways to experiment with the Microsoft API, but they all require Windows as far as I know. In general I think using VBA right from within Word is the easiest method. I used to keep a Windows virtual machine around (on my Mac) to do that sort of thing but don't have it anymore.

Benjamin-T:
Dear @scanny 

Thanks for the response, I've run a test using the Word API:
 [See Docs](https://docs.microsoft.com/en-us/office/vba/api/word.bookmarks.add)
```bash
Sub Mark() 
 ActiveDocument.Bookmarks.Add Name:="mark" 
End Sub
```
## Bookmark at start of document
Results in:

```xml
<w:body xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
  <w:p w:rsidR="000E0575" w:rsidRDefault="000E0575" w14:paraId="6AE70610" w14:textId="77777777" xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml">
    <w:bookmarkStart w:name="myplace" w:id="0" />
    <w:bookmarkEnd w:id="0" />
  </w:p>
  <w:sectPr w:rsidR="000E0575">
    <w:pgSz w:w="11906" w:h="16838" />
    <w:pgMar w:top="1440" w:right="1440" w:bottom="1440" w:left="1440" w:header="708" w:footer="708" w:gutter="0" />
    <w:cols w:space="708" />
    <w:docGrid w:linePitch="360" />
  </w:sectPr>
</w:body>
```

Perhaps interesting, the corresponding generated code: 

```csharp
using DocumentFormat.OpenXml.Wordprocessing;
using DocumentFormat.OpenXml;

namespace GeneratedCode
{
    public class GeneratedClass
    {
        // Creates an Body instance and adds its children.
        public Body GenerateBody()
        {
            Body body1 = new Body();

            Paragraph paragraph1 = new Paragraph(){ RsidParagraphAddition = "000E0575", RsidRunAdditionDefault = "000E0575", ParagraphId = "6AE70610", TextId = "77777777" };
            BookmarkStart bookmarkStart1 = new BookmarkStart(){ Name = "myplace", Id = "0" };
            BookmarkEnd bookmarkEnd1 = new BookmarkEnd(){ Id = "0" };

            paragraph1.Append(bookmarkStart1);
            paragraph1.Append(bookmarkEnd1);
            body1.Append(paragraph1);
            return body1;
        }


    }
}


```

## Bookmark inside a run

I've also placed a bookmark using the same code with the curser inside a run:

```xml
<w:document>
  <w:body>
    <w:p w:rsidR="000E0575" w:rsidRDefault="00390F75" w14:paraId="6AE70610" w14:textId="1AE1D158">
      <w:r>
        <w:t xml:space="preserve">This is a bookmark </w:t>
      </w:r>
      <w:bookmarkStart w:name="myplace" w:id="0" />
      <w:bookmarkEnd w:id="0" />
      <w:r>
        <w:t>inside a run</w:t>
      </w:r>
    </w:p>
    <w:p w:rsidR="00390F75" w:rsidRDefault="00390F75" w14:paraId="6428D04B" w14:textId="77777777" />
    <w:sectPr w:rsidR="00390F75">
      <w:pgSz w:w="11906" w:h="16838" />
      <w:pgMar w:top="1440" w:right="1440" w:bottom="1440" w:left="1440" w:header="708" w:footer="708" w:gutter="0" />
      <w:cols w:space="708" />
      <w:docGrid w:linePitch="360" />
    </w:sectPr>
  </w:body>
</w:document>
```


```csharp
using DocumentFormat.OpenXml.Wordprocessing;
using DocumentFormat.OpenXml;

namespace GeneratedCode
{
    public class GeneratedClass
    {
        // Creates an Body instance and adds its children.
        public Body GenerateBody()
        {
            Body body1 = new Body();

            Paragraph paragraph1 = new Paragraph(){ RsidParagraphAddition = "000E0575", RsidRunAdditionDefault = "00390F75", ParagraphId = "6AE70610", TextId = "1AE1D158" };

            Run run1 = new Run();
            Text text1 = new Text(){ Space = SpaceProcessingModeValues.Preserve };
            text1.Text = "This is a bookmark ";

            run1.Append(text1);
            BookmarkStart bookmarkStart1 = new BookmarkStart(){ Name = "myplace", Id = "0" };
            BookmarkEnd bookmarkEnd1 = new BookmarkEnd(){ Id = "0" };

            Run run2 = new Run();
            Text text2 = new Text();
            text2.Text = "inside a run";

            run2.Append(text2);

            paragraph1.Append(run1);
            paragraph1.Append(bookmarkStart1);
            paragraph1.Append(bookmarkEnd1);
            paragraph1.Append(run2);
            Paragraph paragraph2 = new Paragraph(){ RsidParagraphAddition = "00390F75", RsidRunAdditionDefault = "00390F75", ParagraphId = "6428D04B", TextId = "77777777" };

            body1.Append(paragraph1);
            body1.Append(paragraph2);
            return body1;
        }


    }
}

```

## Bookmark around specific text in run
(And also added a field to test whether the reference is also working correctly)
```xml
<w:body xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
  <w:p w:rsidR="000E0575" w:rsidRDefault="00390F75" w14:paraId="6AE70610" w14:textId="5ABFDEE5" xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml">
    <w:r>
      <w:t xml:space="preserve">This is a </w:t>
    </w:r>
    <w:bookmarkStart w:name="myplace" w:id="0" />
    <w:r>
      <w:t>bookmark inside</w:t>
    </w:r>
    <w:bookmarkEnd w:id="0" />
    <w:r>
      <w:t xml:space="preserve"> a run</w:t>
    </w:r>
  </w:p>
  <w:p w:rsidR="004C3C5C" w:rsidRDefault="004C3C5C" w14:paraId="0DD6D36E" w14:textId="0579ED72" xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml" />
  <w:p w:rsidR="004C3C5C" w:rsidRDefault="004C3C5C" w14:paraId="3196D0A8" w14:textId="17C6E3EF" xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml">
    <w:r>
      <w:fldChar w:fldCharType="begin" />
    </w:r>
    <w:r>
      <w:instrText xml:space="preserve"> REF myplace \h </w:instrText>
    </w:r>
    <w:r>
      <w:fldChar w:fldCharType="separate" />
    </w:r>
    <w:r>
      <w:t>bookmark inside</w:t>
    </w:r>
    <w:r>
      <w:fldChar w:fldCharType="end" />
    </w:r>
  </w:p>
  <w:p w:rsidR="00390F75" w:rsidRDefault="00390F75" w14:paraId="6428D04B" w14:textId="77777777" xmlns:w14="http://schemas.microsoft.com/office/word/2010/wordml" />
  <w:sectPr w:rsidR="00390F75">
    <w:pgSz w:w="11906" w:h="16838" />
    <w:pgMar w:top="1440" w:right="1440" w:bottom="1440" w:left="1440" w:header="708" w:footer="708" w:gutter="0" />
    <w:cols w:space="708" />
    <w:docGrid w:linePitch="360" />
  </w:sectPr>
</w:body>
```

macpoupou:
Have you plan to merge it soon ?

scanny:
Closing for now, may come back to this later.

Issue #232, opened 2015-11-30
How to detect merged cells when reading tables


sylvain-bougnoux:
Hello, I've created a simple .docx document (Word 2010) with a simple 2x2 table, the 1st col is merged as with:
+---+---+
|      | b  |
+ a +---+
|      | c  |
+---+---+
When reading this with python the cells (0,0) and (1,0) are different!
How can I detect that they are merged on the original document?
The text of both cell is indeed 'a'. But the 'pointer' are different.
Thanks
Python 2.7, docx 0.85

sylvain-bougnoux:
I've found a workaround (though the doc does not stand).
With the above table in simple.docx

```
from docx import Document
doc=Document("simple.docx")
table=doc.tables[0]
c00=table.cell(0,0)
c10=table.cell(1,0)
c00==c01  # is false as reported
#however
c00._tc==c01._tc # is true
```
Regards

oltish:
It has been a while, but someone else may need this information in future.

As mentioned in previous comment, docx.table._Cell object has property _tc which contains useful information about cell's span:

```
from docx import Document
doc=Document("simple.docx")
table=doc.tables[0]

c=table.cell(0,0)
print(c.text, c._tc.top, c._tc.bottom, c._tc.left, c._tc.right)

c=table.cell(0,1)
print(c.text, c._tc.top, c._tc.bottom, c._tc.left, c._tc.right)

c=table.cell(1,1)
print(c.text, c._tc.top, c._tc.bottom, c._tc.left, c._tc.right)

```
The output will be:
```
a 0 2 0 1
b 0 1 1 2
c 1 2 1 2
```

From this you will easily understand not only if the cell is merged or not, but also its shape and size.

0-173:
I found that some docx files created by word have inconsistend grid_span attributes in a row.
When accessing a specific cell the ._cells property will not care about the rows defined by the xml but  just count through the cells and assume that the cell count per row and all grid_span are correct.
In my case there were 7 columns in the table but one row contained only grid_span attribues of 1 + 1 + 2 + 2 = 6. After this row, all the succeeding rows are broken.
The workaround for detecting this has already been mentioned by @oltish.
I will provide a merge request as a fix for the _cells function.

mrlnc:
The workaround from @0-173 works for me. Thanks! However, the new cell doesn't have any attributes like the "text" field.

fusted2:
> c00._tc==c01._tc # is true

[...]

PR #564, opened 2018-10-29
detect beginning of new row. if last row is incomplete, add empty cells.

0-173:
I found that some tables in MS-Word created documents have inconsistend grid_span attributes in a row (not matching the tblGrid's gridCol element count).

When accessing the ._cells property of such a table, it will not consider the row/column definition of the xml but just count the cells and assume that the cell count per row and all grid_span attributes are consistent with the tblGrid specification.
In my case there were 7 columns in the table but one row contained only grid_span attribues of 1 + 1 + 2 + 2 = 6. The next row and all succeeding rows would get partly wrapped around and were broken.

This merge request fixes the issue by detecting a row's start (see #232) and then filling any incomplete row with empty cells if necessary. Unit test is included.

### DIFF
--- a/docx/table.py
+++ b/docx/table.py
@@ -168,6 +168,11 @@ class Table(Parented):
         col_count = self._column_count
         cells = []
         for tc in self._tbl.iter_tcs():
+            previousRowIncomplete = (tc.left == 0) and (len(cells) % col_count > 0)
+            if previousRowIncomplete:
+                missingCellCount = col_count - (len(cells) % col_count)
+                for i in range(0, missingCellCount):
+                    cells.append(_Cell("w:tc", self))
             for grid_span_idx in range(tc.grid_span):
                 if tc.vMerge == ST_Merge.CONTINUE:
                     cells.append(cells[-col_count])
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -69,6 +69,11 @@ class DescribeTable(object):
         row_cells = table.row_cells(row_idx)
         assert row_cells == expected_cells
 
+    def it_detects_the_beginning_of_new_rows(self, inconsistet_col_span_fixture):
+        table, row_idx, col_idx, expected_text = inconsistet_col_span_fixture
+        cell = table.cell(row_idx, col_idx)
+        assert cell.text == expected_text
+
     def it_knows_its_alignment_setting(self, alignment_get_fixture):
         table, expected_value = alignment_get_fixture
         assert table.alignment == expected_value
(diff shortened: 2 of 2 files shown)


scanny:
Can you say how the documents were created and how one with this characteristic can be reproduced?

0-173:
Hi Scanny, thats a good question. I currently only have a document that has been generated in MS Word by our customer. There are multiple people involved in the creation of this document, so it's not so easy to find out how exactly they produced this document - but I will give it a try. I have modified the original document to obtain a minimal version demonstrating the problem. The table has a gridCol definition of 7 columns. In this table the first line consists of cells that sum up to 7 columns, the second and third line each sum up to only 6 columns.
[84_minimal_demo.docx](https://github.com/python-openxml/python-docx/files/2529640/84_minimal_demo.docx)

0-173:
I talked once again to the customer. They're copy-pasting lot of cells and rows from tables in other documents. But we are still not able to reproduce such a document from scratch.
Do you think it makes sense to integrate this new-row-detection? In my opinion, even though the document is kind of corrupted, the behaviour of not-seeing the new row (which is defined in the xml) is not what we would expect from the library.

aidanfindlater:
I'd just like to mention that I found this patch helpful.  I'm having a similar problem with unwieldy manually-edited-over-years tables.  The original document is from 2012, so it has acquired from cruft over the years.  Some of the tables have different row lengths (for aesthetics, I believe), or cell widths that have been adjusted independently of the column (which has the same effect on the script, from what I can tell).

The expected behaviour is for the library to produce essentially an m x n dataframe that corresponds as closely as possible to the original table.  The most straightforward way to do this is the one suggested by @0-173, where the shortened row would have empty cells added to the end.  Instead, every time the script reached one of these inconsistencies, it fills any missing cells in the shortened row with cells from the next row (removing them from the next row, and thereby offsetting the data in the rest of the dataframe).  Essentially, its current behaviour is to make an n x m table, then fill it in with whichever cells happen to exist (going LTR, TTB), regardless of which location they were originally in.

I've got it working now my monkey-patching my script with a version of this patch, so thank you.

blackdaemon:
This is happening to me after converting doc to docx using Microsoft Office Migration Planning Manager. Resulting docx looks the same as the source doc visually, but internally it exhibits this issue.

scanny:
Ah, that's a great datapoint @blackdaemon, it makes sense that the document was operated on by something other than Word itself. Thanks for mentioning that :)

fdabek1:
@scanny Is this going to get merged in and be released?  I have come across this same issue with a bunch of files and my only solution at the moment would be to delete the extraneous column manually.  Getting this merged would be a huge help!

HiGregSmith:
I wonder if this issue is related to the gridAfter and gridBefore properties. I have noticed in the table I'm working with that a fewer number of tcPr elements happens when trPr also has a gridAfter element defined for the row as non-zero. If either gridAfter or gridBefore are unspecified, zero is the default..

It seems to be the case that the the table's number of columns equals, for each row:
gridAfter (row property, zero if unspecified), plus
gridBefore (row property, zero if unspecified), plus
each cell's gridSpan property (one in each cell where it is unspecified).

For more details, see https://github.com/python-openxml/python-docx/issues/881

FredEPr:
It's actually really easy to produce a table with a mismatch column count on Word. 

On any table you can right click a cell and (I don't have Word in English so I'm guessing the names) insert -> insert cells... -> shift cells to the right. It will insert a single cell and shift the remaining cells of the row to the right. Or you can do the exact opposite and deleting a single cell with right click -> delete cells... -> shift cells to the left. 

Both will produce a table with an inconsistent number of columns and accessing the table via row and column indexes will be unreliable.

The patch from 0-173 fix this issue.

By the way the issue #861 is probably the same.

tonal:
Try https://github.com/python-openxml/python-docx/issues/881#issuecomment-1437979336

PR #571, opened 2018-11-12
WIP: Feature Header & Footer

ondrej-111:
implemented header and footer by spec but without actions in: https://github.com/python-openxml/python-docx/blob/master/docs/dev/analysis/features/header.rst

linked to issue: https://github.com/python-openxml/python-docx/issues/104

in progress: implement add/remove_header/footer


Benjamin-T:
@ondrej-111  first of all: Nice work! 

I've looked briefly at your code and I've noticed that you chose to create two complex types: one for the header and one for the footer. When I was investigating the parts - complex types for the Bookmarks (PR #445 ) I found that the header and footer share the same comlex type: `CT_HdrFtr`. This make sense as you expect identical behavior for the header and the footer. 

see `wml.xsd`:
https://github.com/python-openxml/python-docx/blob/94623ff627f400991c9e550a9da7473ef3acf36f/ref/xsd/wml.xsd#L3441-L3442

which would mean that:
https://github.com/python-openxml/python-docx/blob/99c1e323569dbfff88ecbbbdd7651ec95879a83c/docx/oxml/__init__.py#L79-L80
should refer to CT_HdrFtr

And I think it's required that in order to get a PR merged, It should start with the `analysis document` based on this document, create an acceptance test. 

When this acceptance test is run using `behave --stop` it will fail at some missing code; this code should be added as next commit. 

However, I would first try and create some working version a `spike` and based on this continue towards the TDD stage where each separate method appears in a separate commit together with its unit tests.

regards,

Ben

ondrej-111:
@Benjamin-T Thank You for your help.

Analysis document already exist here:
[/dev/analysis/features/header.rst](https://github.com/python-openxml/python-docx/blob/master/docs/dev/analysis/features/header.rst).

I think I can close this PR and create new branch like feature/header_footer with right work flow where I will start with acceptance tests and then implement code and unit tests step by step...

PR #577, opened 2018-11-20
Table: feature : allow table looking (header row/col, footer row/col, bands) modification

lnv42:
I added a feature needed for another project that allow modification of table looking parameters.
It permit to set:
first/last row/col and horizontal/vertical bands values for each table at the table creation or later.

I believe that it might be useful to someone else so if you agree with that: you can merge it.


lnv42:
Sorry i forgot to PR in my feature branch and i did a PR of all my master, i'll submit a proper PR.

PR #587, opened 2018-12-15
Fix linter warning for Python 3.8+ support for collections

tony:
Fixes warning that comes up with pytest. Also fixes Python 3.8 forward compatibility.

```
.venv3/lib/python3.7/site-packages/docx/section.py:9
  DeprecationWarning: Using or importing the ABCs from 'collections'
  instead of from 'collections.abc' is deprecated, and in 3.8 it will
  stop working
```

See also: https://github.com/pallets/jinja/commit/31bf9b7e71c, https://github.com/python/cpython/commit/c66f9f8d

### DIFF
--- a/docx/compat.py
+++ b/docx/compat.py
@@ -10,6 +10,11 @@ from __future__ import (
 
 import sys
 
+try:  # Python 3.8+
+    from collections import abc
+except ImportError:
+    import collections as abc
+
 # ===========================================================================
 # Python 3 versions
 # ===========================================================================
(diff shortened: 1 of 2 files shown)


tony:
@scanny rebased

watfordgnf:
@scanny How can we help to get this moving forward? This will block our adoption of Python 3.9 and we'd rather not fork the project.

tirkarthi:
This seems to have been fixed in https://github.com/python-openxml/python-docx/commit/da75fcf01f7f322e846e2ac3e1936aedd766acc8 . See also https://github.com/python-openxml/python-docx/pull/681.

> Fixed in release v0.8.11 on May 15, 2021.

scanny:
Ah, right, good catch @tirkarthi, closing this one up.

PR #628, opened 2019-03-08
Add testing and document support for Python 3.7

jdufresne:
Python 3.7 was released on June 27, 2018.

https://docs.python.org/3/whatsnew/3.7.html


jdufresne:
This now includes Python 3.8 as well. Python 3.8 was released on October 14th, 2019.

PR #636, opened 2019-03-08
Fix deprecation warning by importing from collections.abc

### LINKED ISSUE #619: Warning: Using or importing ABCs from collections is deprecated
I recently encountered the following warning while running `python-docx`:

```
section.py:7: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

I noticed that other third-party packages have encountered this same issue and resolved it. See https://github.com/marshmallow-code/marshmallow/issues/1027 for example.

jdufresne:
Fixes warning of the form:

    section.py:7: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

Fixes #619

### DIFF
--- a/docx/section.py
+++ b/docx/section.py
@@ -4,7 +4,11 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from collections import Sequence
+try:
+    from collections.abc import Sequence
+except ImportError:
+    # Python 2.7
+    from collections import Sequence
 
 from docx.blkcntnr import BlockItemContainer
 from docx.enum.section import WD_HEADER_FOOTER


jdufresne:
Looks like this was fixed in da75fcf01f7f322e846e2ac3e1936aedd766acc8

scanny:
Yes, but not released yet.

jdufresne:
But why have a 2nd PR to duplicate the work?

scanny:
Yeah, no, you're right; I was thinking I needed to remind myself (I typically retire the `shortlist` labeled items on a release) but since this one is already on master, I expect it will be hard to forget :)

Issue #284, opened 2016-03-21
Add docm support


blindidiot91:
This module is great, but it doesn't currently work with docm files or macros. 

Add in a way to both import macros into the VBProject and save docm files. 

Currently, the only way to do this is pywin32 and if you both import a macro and save as docm for 5 files, it takes about 30 sec-1 minute which is obviously unacceptable.

Your module is perfectly positioned to take care of this problem and I'd hope to see an improvement sometime in the future.

Thanks!

scanny:
Do you have an example of a .docm file you can share? (You can just drag and drop it in the comment box in response.)

I had understood that macros were stored in a binary format, so I'm not sure how you'd be able to generate them in Python, even if you could "import" them.

But I'll take a look at the file structure and see what we see.

ethack:
I would like to see this implemented as well.  I've attached a sample document containing a Hello World macro.

Since Github wouldn't let me upload the document as-is, I changed the extension to ".zip".

[MacroTest.docm.zip](https://github.com/python-openxml/python-docx/files/185043/MacroTest.docm.zip)

For reference, here is the macro I put in the document:
```vba
Sub AutoOpen()
    MsgBox ("Hello, World!")
End Sub
```

scanny:
Thanks @ethack, just the ticket :)

Here's what the innards look like:
```bash
$ unzip -l MacroTest.docm                                                                                                                            Archive:  MacroTest.docm
  Length     Date   Time    Name
 --------    ----   ----    ----
     2485  01-01-80 00:00   [Content_Types].xml
      590  01-01-80 00:00   _rels/.rels
     1976  01-01-80 00:00   word/_rels/document.xml.rels
     1977  01-01-80 00:00   word/document.xml
     1295  01-01-80 00:00   word/header3.xml
     1295  01-01-80 00:00   word/footer2.xml
     1295  01-01-80 00:00   word/footer1.xml
     1295  01-01-80 00:00   word/header2.xml
     1295  01-01-80 00:00   word/header1.xml
     1675  01-01-80 00:00   word/endnotes.xml
     1681  01-01-80 00:00   word/footnotes.xml
     1295  01-01-80 00:00   word/footer3.xml
     6795  01-01-80 00:00   word/theme/theme1.xml
      277  01-01-80 00:00   word/_rels/vbaProject.bin.rels
     9728  01-01-80 00:00   word/vbaProject.bin
     2699  01-01-80 00:00   word/settings.xml
[...]

PR #673, opened 2019-05-13
Add '.docm' compatibility

DylanGnd:
Hello, I saw few topics about '.docm' (docx with maccro) compatibility files so I decided to create a pull request. I think that could help some peoples.

### DIFF
--- a/docx/__init__.py
+++ b/docx/__init__.py
@@ -33,6 +33,8 @@ PartFactory.part_type_for[CT.WML_HEADER] = HeaderPart
 PartFactory.part_type_for[CT.WML_NUMBERING] = NumberingPart
 PartFactory.part_type_for[CT.WML_SETTINGS] = SettingsPart
 PartFactory.part_type_for[CT.WML_STYLES] = StylesPart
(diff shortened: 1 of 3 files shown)


jshlbrd:
@scanny any chance we can get a review on this? Related to this 3yo issue https://github.com/python-openxml/python-docx/issues/284

lordofhyphens (approved):
Looks pretty straightforward to me for minimal additional work.

goncalor:
Hi. Will this PR not get merged? Thank you.

scanny:
Not in its present state. There is no analysis documentation and no tests. The analysis document would need to account for how macros are stored in the `.pptx` package and document experiments that showed the macros were preserved on save and otherwise support reasoning about the mechanism used to establish support for `.docm` files and where it may be insufficient or have unintended consequences.

If this solution is good enough for your purposes you can just merge it into a private fork of the latest release and use that private fork.

tsuga:
FYI, https://github.com/python-openxml/python-docx/pull/716

PR #681, opened 2019-06-03
import Sequence from collections.abc

itrevex:
Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

### DIFF
--- a/docx/section.py
+++ b/docx/section.py
@@ -4,7 +4,7 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from collections import Sequence
+from collections.abc import Sequence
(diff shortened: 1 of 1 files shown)


brasky:
Looks like this breaks python 2.x compatibility? Could do something like this instead:
```python
try:
    from collections.abc import Sequence
except ImportError:
```

jdufresne:
Looks like this was fixed in da75fcf01f7f322e846e2ac3e1936aedd766acc8

scanny:
Fixed in release v0.8.11 on May 15, 2021.

PR #691, opened 2019-07-03
Prevent "invalid literal for int() with base 10" when converting

danielhavir:
Prevent \"invalid literal for int() with base 10\" when converting float str_value to int.


charanjit-singh:
Why was this not merged?

danielhavir:
The PR was open for 9 months with no response, so I eventually closed it.

PR #697, opened 2019-07-15
Develop

BayooG:
Hi @scanny! 
In the last few months, I've been working on a project and everything was fine until I needed things that are not implemented in python-docx so I forked it and added the things I needed
in this branch, I added a full implementation for  comments & footnotes (on OXML level and high level)
I also added an implementation for the missing things in the numbering part, added support for emf images, added a high-level implementation to know a table section and finally added an element property to the document object so you can iterate all document Paragraphs, Tables and Section in the right order
Please review this branch and let me know if we can work this out


westscz:
Is there any chance that PR will be merged and package will be delivered soon? Can I help somehow?

PR #798, opened 2020-03-18
Some hack to include SVG

pfernique:
`shape.py` and `pkgwriter.py` are particularly ugly but it works with Word for Office 365 (16.0)

### DIFF
--- a/docx/oxml/shape.py
+++ b/docx/oxml/shape.py
@@ -24,7 +24,6 @@ class CT_Blip(BaseOxmlElement):
     embed = OptionalAttribute('r:embed', ST_RelationshipId)
     link = OptionalAttribute('r:link', ST_RelationshipId)
 
-
 class CT_BlipFillProperties(BaseOxmlElement):
     """
     ``<pic:blipFill>`` element, specifies picture properties
@@ -81,7 +80,10 @@ class CT_Inline(BaseOxmlElement):
         specified by the argument values.
         """
         pic_id = 0  # Word doesn't seem to use this, but does not omit it
-        pic = CT_Picture.new(pic_id, filename, rId, cx, cy)
+        if filename.endswith('.svg'):
+            pic = CT_SVGPicture.new(pic_id, filename, rId, cx, cy)
+        else:
+            pic = CT_Picture.new(pic_id, filename, rId, cx, cy)
         inline = cls.new(cx, cy, shape_id, pic)
         inline.graphic.graphicData._insert_pic(pic)
         return inline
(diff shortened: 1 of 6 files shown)


mrsanrise:
Have tried this(https://github.com/pfernique/python-docx/tree/feature/add_svg_picture) with word 2013.
"This image cannot currently be displayed". ((

pfernique:
Unfortunatly SVG works only on word >= 2016 (more details [here](https://superuser.com/questions/397644/inserting-svg-files-in-a-microsoft-word-document))

mrsanrise:
> Unfortunatly SVG works only on word >= 2016 (more details [here](https://superuser.com/questions/397644/inserting-svg-files-in-a-microsoft-word-document))

Thank you for the link. I think that is not very important which vector format to use.
I have converted all my images to the wmf (my images source format is odg).
After that i wrote my `class Wmf(BaseImageHeader)`. It is similar to your example with svg (Thank you! :) ).  I got acceptable result - word shows these images . Now i have a small issue with a scaling for some images (but in my case it is not critical).
Example function for the WMF support:

```python
        #https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/828e1864-7fe7-42d8-ab0a-1de161b32f27
        stream.seek(0)
        META_PLACEABLE = stream.read()
        print("META_PLACEABLE")
        mkey = META_PLACEABLE[0:4]  # Key (4 bytes): Identification value that indicates the presence of a placeable metafile header. This value MUST be 0x9AC6CDD7.
        mhmw = META_PLACEABLE[4:6]  # HWmf (2 bytes): The resource handle to the metafile, when the metafile is in memory. When the metafile is on disk, this field MUST  contain 0x0000. This attribute of the metafile is specified in the Type field of the META_HEADER Record.
        mbbox = META_PLACEABLE[6:14] # BoundingBox (8 bytes): The rectangle in the playback context (or simply the destination rectangle), measured in logical units, for displaying the metafile. The size of a logical unit is specified by the Inch field.
        minch = META_PLACEABLE[14:16] # Inch (2 bytes): The number of logical units per inch used to represent the image. This value can be used to scale an image.
       
```

jayceslesar:
Merge at some point?

tobytraylor:
Would like to see this merged.

docx/oxml/shape.py
```diff
@@ -81,7 +80,10 @@ def new_pic_inline(cls, shape_id, rId, filename, cx, cy):
         specified by the argument values.
         """
         pic_id = 0  # Word doesn't seem to use this, but does not omit it
-        pic = CT_Picture.new(pic_id, filename, rId, cx, cy)
+        if filename.endswith('.svg'):
```
awtimmering, line 83: Unless I'm misreading this - this would only capture SVG is actually provided as a file (that has a `filename` go with it). There are many use cases where pictures are generated on the fly in in-memory buffers and therefore not provided as files. Any way to account for that? 
(For example the PNG equivalent for matplotlib + docx would be something like:
```
import base64
import io

import docx
import matplotlib.figure as figure

fig = figure.Figure()
ax = fig.subplots()
ax.plot([1, 2])
buf = io.BytesIO()
fig.savefig(buf, format='png')

document = docx.Document()
document.add_heading('Document Title', 0)
document.add_picture(buf, width=Inches(1.25))
document.save('demo.docx')
```

pfernique: That's true that this seems strange but the when dispatch is done, a filename `image.svg` is given for a stream (must be the same for `png`). So even if you provide in-memory buffers, it's good.

awtimmering: Didn't realize that - thank you for calling it out. I looked a bit further (which I should have from the beginning):filename is generated [here](https://github.com/python-openxml/python-docx/blob/da75fcf01f7f322e846e2ac3e1936aedd766acc8/docx/image/image.py#L177) in `image.py`.

mrgou:
Definitely looking forward to this. Word now support SVG _natively_, i.e. images are imported as editable vector objects.

PR #908, opened 2020-12-30
Add support for RGBColor font highlight

stdedos:
* Define `w:shd` empirically (based on `w:highlight`)
  So far, `w:shd` and `w:highlight` appear to be the same attributes
  for different values (`RGBColor` vs `WS_COLOR_INDEX`)
* Add `{,.}venv` on `.gitignore`
* Add `EnumValue.name` attribute, to be able to retrieve an attribute's name
* Fix a typo

Signed-off-by: Stavros Ntentos <133706+stdedos@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 _scratch/
 Session.vim
 /.tox/
+.venv/
+venv/
--- a/docx/enum/base.py
+++ b/docx/enum/base.py
@@ -287,6 +287,10 @@ class EnumValue(int):
         self._docstring = docstring
 
     @property
+    def name(self):
+        return self._member_name
+
+    @property
     def __doc__(self):
         """
         The description of this enumeration member
--- a/docx/oxml/__init__.py
+++ b/docx/oxml/__init__.py
@@ -185,6 +185,7 @@ from .text.font import (  # noqa
     CT_Color,
     CT_Fonts,
     CT_Highlight,
+    CT_Shd,
     CT_HpsMeasure,
     CT_RPr,
     CT_Underline,
@@ -198,6 +199,7 @@ register_element_cls('w:cs',         CT_OnOff)
 register_element_cls('w:dstrike',    CT_OnOff)
 register_element_cls('w:emboss',     CT_OnOff)
 register_element_cls('w:highlight',  CT_Highlight)
+register_element_cls('w:shd',        CT_Shd)
 register_element_cls('w:i',          CT_OnOff)
 register_element_cls('w:iCs',        CT_OnOff)
 register_element_cls('w:imprint',    CT_OnOff)
--- a/docx/oxml/text/font.py
+++ b/docx/oxml/text/font.py
@@ -14,6 +14,7 @@ from ..simpletypes import (
 from ..xmlchemy import (
     BaseOxmlElement, OptionalAttribute, RequiredAttribute, ZeroOrOne
 )
+from ...shared import RGBColor
 
 
 class CT_Color(BaseOxmlElement):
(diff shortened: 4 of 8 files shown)


stdedos:
While this is a self-sufficient commit for its work, this opens up aeolian bags:

Defining `CT_Shd` is incomplete, which requires:
```patch
diff --git a/docx/oxml/text/font.py b/docx/oxml/text/font.py
index 9c7f90b..d83e2ca 100644
--- a/docx/oxml/text/font.py
+++ b/docx/oxml/text/font.py
@@ -46,7 +46,9 @@ class CT_Shd(BaseOxmlElement):
     """
     `w:shd` element, specifying font highlighting/background color.
     """
+    color = RequiredAttribute('w:color', ST_HexColor)
     fill = RequiredAttribute('w:fill', ST_HexColor)
+    val = RequiredAttribute('w:val', ST_Shd)
 
 
 class CT_HpsMeasure(BaseOxmlElement):
```

Which requires defining `ST_Shd` (23 attributes, http://www.ecma.ch/activities/Office%20Open%20XML%20Formats/Draft%20ECMA-376%203rd%20edition,%20March%202011/Office%20Open%20XML%20Part%204%20-%20Transitional%20Migration%20Features.pdf, page 929). LibreOffice defines `w:val=clean` when using just highlight; but it doesn't seem to be necessary (opening the document without it looks okay).

Defining them, makes them setting even harder - since `highlight.val` is all about an `Enum`, and `shd` controls more attributes (which cannot be hidden behind just `highlight_color`).

It may also require further deciphering what is required and what is not on partial definitions.

----

Defining `shd` and `highlight` as "features with parity", means that they interact with each other (setting one unsets the other, getting values requires checking both, clearing also), which makes their code tangle.

----

Finally, while I think my tests are well-defined - they fail:

```python
tests/text/test_font.py:83 (DescribeFont.it_knows_its_highlight_color[highlight_get_fixture4])
self = <tests.text.test_font.DescribeFont object at 0x7f0ed6c58d68>
highlight_get_fixture = (<docx.text.font.Font object at 0x7f0ed7146470>, RGBColor(0x61, 0x95, 0xed))

    def it_knows_its_highlight_color(self, highlight_get_fixture):
        font, expected_value = highlight_get_fixture
>       assert font.highlight_color is expected_value
E       assert RGBColor(0x61, 0x95, 0xed) is RGBColor(0x61, 0x95, 0xed)
E        +  where RGBColor(0x61, 0x95, 0xed) = <docx.text.font.Font object at 0x7f0ed7146470>.highlight_color

tests/text/test_font.py:86: AssertionError
```

Of course, I think the `is` is to blame; but I am not sure why is the `is` mandatory, and a simple `==` won't do.

scanny:
On the test, all prior values were singletons, so `is` was appropriate, if perhaps a little overkill. Since the new value you introduced is _not_ the same object as the (separately created) expected value, `is` would not be appropriate and would be changed to `==`, just as you say.

The only question in my mind is whether that diversity in return types is appropriate given the difficulty it introduces in _using_ the interface. If you don't know the type of the return value, you need a fair amount of code on every call just to discern how to handle the result.

The `Font.highlight_color` interface is based on the `Range.HighlightColorIndex` interface in the Microsoft Word API here: https://docs.microsoft.com/en-us/office/vba/api/word.range.highlightcolorindex, and the existing implementation seems pretty complete as far as that interface is concerned.

I would be inclined to find the (separate I expect) MS-API interface that allows the (differently implemented but similar effect) shading you are after; I think it is this one: https://docs.microsoft.com/en-us/office/vba/api/word.font.shading. Note that it (always) returns a `Shading` object and thereby presents a consistent interface to the caller.

We're probably not going to implement this soon, so don't feel obliged to change your implementation on our account. As long as it works for your use case it's probably not going to matter much.

We do appreciate your pull request and research though, and would like to keep this pull request as a basis for possibly implementing this feature later.

stdedos:
Would you consider the other small changes for inclusion?

I can (_or someone else can_) graft them out and push them if so.

stdedos:
> The only question in my mind is whether that diversity in return types is appropriate given the difficulty it introduces in _using_ the interface. If you don't know the type of the return value, you need a fair amount of code on every call just to discern how to handle the result.

I thought about this; I have sided with the "usability" side of the question (hence this code).

I don't know if it's just me, the documentation, "some lack of examples", any, all or none of the above; for me, there was too much friction in trying to use the library to create a simple (~200 LOC) x-to-y converter. Therefore, I wanted to make the implementation "as stupid as possible".

Issue #685, opened 2019-06-16
how to make the first line by 2 ch?


paopjian:
paragraph_format.first_line_indent can only set by direct numbers, in word for Chinese,  the paragraph indentation special can set First line by 2 ch. I don't know how to make it.
I find CharacterUnitFirstLineIndent in VBA,but not find in python-docx
I find that w:firstLineChars can make "by" show the ch

![image](https://user-images.githubusercontent.com/20377352/59559791-f9354980-903b-11e9-9fb2-61a96a22b932.png)
![image](https://user-images.githubusercontent.com/20377352/59559794-05210b80-903c-11e9-89be-dfefc9e8acbc.png)

gc13141112:
The configuration  "paragraph_format.first_line_chars_indent = 2 "  is not working. How to set the scripts?

PR #941, opened 2021-05-05
add w:firstLineChars in text for #685 fixed

listeng:


etp30:
hi, great work. but, Is it a better choice to read/write hangingChars instead of hanging in this property, for the purpose of avoiding ambiguity with first_line_indent property?

PR #958, opened 2021-06-11
Fix "KeyError: no style with name 'Heading 1'" when using a file edit…

AndresRReina:
…ed with Libre Office.

This error happens due to the style "Heading 1" being capitalized in the docx xml, unlike MS Word does.


kugelpunk:
@cristianMuntean1 your comment about your [CLI app](https://github.com/python-openxml/python-docx/pull/958#pullrequestreview-790383517) has accidentally been pasted here. Please, remove it.

Issue #413, opened 2017-07-06
Accessing the content of TextBox and Headers (for purpose of find and replace)


bh2smith:
I am hoping to do a find and replace and have managed to perform this quite well with Paragraphs and Tables, but I am still not able to determine how to do the same with TextBox and Headers. Is there any way to access this content with the current version?

scanny:
Short answer is no. There's no API support for accessing header/footers nor is there any for "floating" text boxes.

There is some work-in-progress on headers, but it seems to have stalled.

SwimSweet:
Can you give any way to modify the textbox?thx.

MatthewP-Newbee:
Hey, check weather this [website](https://documentation.devexpress.com/OfficeFileAPI/116845/Word-Processing-Document-API/Examples/Text-Boxes/How-to-Insert-Text-Box) useful or not.

scanny:
Headers and footers were added a while back. Closing as resolved.

bh2smith:
I don't even remember what I was trying to do here... but thanks for resolving it!

PR #960, opened 2021-06-13
Add support for editing textbox content

### LINKED ISSUE #524: Writing to a Textbox.
Is there a way currently to create and write into a text box within a word file using python-docx?
I am looking to enter formatted text directly into a text box and potentially even position it at a specific location on the page. Ideally I would like to have the text box have no fill or border as well.

RikVoorhaar:
fixes #524, partially addresses #413 

Creating text boxes is a little complicated, the main problem is current lack of support for floating shapes. However, editing text boxes is not very hard to implement. My personal use case is the need to create word documents from a template, and templates often contain text boxes.

I have no prior experience working with this package, or XML / word documents in general, so I will need some help with this.

## Analysis

It seems that in the OXML standard textboxes always appear inside an ``<mc:AlternateContent>`` element. The reason is that text boxes (and shapes in general) were originally implemented using VML, but since word 2010 it preferentially uses drawingML. This means that now the XML always stores two identical copies of the text box, but for legacy reasons uses two different methods of storing the shape. 
Therefore I define a textbox by an ``<mc:AlternateContent>`` element with two descendant ``<w:txbxContent>`` elements with an xpath as follow:
```
XPATH_CHOICE = "./mc:Choice/w:drawing/wp:anchor/a:graphic/a:graphicData/wps:wsp/wps:txbx/w:txbxContent"
XPATH_FALLBACK = "./mc:Fallback/w:pict/v:*/v:textbox/w:txbxContent"
```
## Remarks

- Whenever we edit a text box, we need to update these two elements in the exact same way. This is managed by my `TextBox` class, and the user should never have to care about this.

- In order to get this to work I had to add namespaces `"mc"`, `"v"`, and `"wps"` to `ns.py`. 

- It seems text boxes are stored in the first run of the first paragraph of the document, at least for one-page documents. In order to make text boxes easy to access I added a `textboxes` property to the `Document` class that recursively searches for all text boxes in the entire document. This is probably not the best way to do this, but I can't think of a better way.

- I haven't written any tests yet, but at least it works fine for my personal usecase. Since this doesn't support creation of text boxes, but only editing existing ones, we need to include a test document to edit. Not 100% what the best way to go about that is.

- Docs also need to be updated


scanny:
It's generally well worthwhile to determine how the Microsoft Word API accomplishes the task and model after that. The few times I didn't early on turned out to be very costly in terms of rework.

Generally I start with a Google search on "vba word {X}", like "how do I do this in Visual Basic" and that will get me into the right neck of the woods. From there I can find the details from the MS Word API documentation: http://msdn.microsoft.com/en-us/library/office/ff837519.aspx.

Based on a quick look, it appears that a textbox is a `Shape`, like a picture or chart is. In the MS API there is a `Document.shapes` property that returns this collection. Probably that's what you would want to do here, where the shapes collection would contain _all_ the shapes ("floating" objects on the "drawing layer" of the document), not just the textboxen.

You'll want to understand the difference between the (flowed) "story layer" (aka. document layer, text layer) and the arbitrarily positioned "drawing layer". All "floating" things are in the drawing layer. The existing "inline" picture shapes (like `run.add_picture()` are in the story layer.

RikVoorhaar:
Sorry for the slow reply -- busy period. 

What you're saying makes sense, what I did was very much hacked together. What I did works for my personal needs, and I unfortunately don't  have much time to dig much deeper into this. 

I'll close the PR. If anyone wants to work on this, perhaps my work can be a foundation.

Issue #455, opened 2018-01-12
add_picture corruptes file in some case


Tores76:
Dear @scanny ,

First of all, thanks to you and all contributors for python-docx!

I have found a problem when I wanted to add images in an existing file. After the insertion the file resulted corrupted when opening it with MS Word.

I did some debugging I have found out that it was a clash between the id of the wp:docPr of the newly inserted image and the wp:docPr of a shape in the header.

I think that the problem is in next_id(self) of document.py that from my understanding looks only for the image ids in document.xml and not in header1.xml (where the conflicting wp:docPr is).
I have temporarily fixed it for my case by artificially increasing the id number of the newly inserted pictures by 10000 as I haven't find out as to access the header1.xml elements and I'm very short on time.

Again, thanks for this extremely useful package!

scanny:
Thanks @Tores76. This is an interesting and useful finding. Your hypothesis makes sense to me; Word has access to the header whenever it has the document and I suppose it's reasonable they would place those ids into the same scope of uniqueness.

We don't have programmatic access to the header XML document(s) yet, so we wouldn't be able to search there yet. The artificial increase in ID for new ones we add makes sense to me though. I'll bet that's a pretty robust approach that would work in all cases we're likely to encounter.

I'll leave this open and put it on the short list for things we can fix next time we're in there.

scanny:
@Tores76 By the way, can you post the code you used to increase the ID on new images you inserted? That could be very helpful to someone who needs it until we can get this into the build.

Tores76:
Sure!
The solution I used at first was some kind of a one shot brute force in the code itself to check my hypothesis. Something like this:

from docx import Document
from docx.oxml.shared import qn

my_document = Document('without_images.docx')

# code inserting some images

doc_element = my_document._part._element
docPrs = doc_element.findall('.//' + qn('wp:docPr'))
for docPr in docPrs:
    docPr.set('id',str(int(docPr.get('id'))+100000))

my_document.save('with_images.docx')

And the output file was finally ok without having to remove the images in the header. A quick test with opc-diag seemed to show that when adding images with Word later on, it keeps using the lower unused ids without worrying about these high value ones.
This solution is straightforward and simple and may be all that is needed (it was for me) but is of course thought to be done once only at the end of all the insertions (the pool of ids could be considered messy if that step was to be repeated multiple times).

Now however I'm working with a modified property next_id in the document.py file:

    @property
    def next_id(self):
        """
[...]

PR #1024, opened 2021-11-03
Fix next_id to support multiple pictures

vbeland:
Implementing the fix for multiple pictures as documented in this issue:
    https://github.com/python-openxml/python-docx/issues/455

### DIFF
--- a/docx/parts/story.py
+++ b/docx/parts/story.py
@@ -60,17 +60,16 @@ class BaseStoryPart(XmlPart):
 
     @property
     def next_id(self):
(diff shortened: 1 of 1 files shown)


vbeland:
@scanny I'm not sure how to assign you as a reviewer.

janpieter-vanparys:
Has there been any update on this issue? We are having this problem but we can't implement this fix manually with every new deploy.

janpieter-vanparys:
@scanny could you implement the fix from @vbeland we could really use it

PR #1036, opened 2021-12-22
Add function can read merged cells

znsoooo:
return current table's merged cells' info sort as [(r1, r2, c1, c2), ...].


znsoooo:
**Usage like:**

```python
import docx
doc = docx.Document('example.docx')
for table in doc.tables:
    print(table.merges)  #  return: [(r1, r2, c1, c2), ...]
```

PR #1051, opened 2022-02-16
added cache for table cells

samzhangjy:
Hi.
I added a cache when accessing table cells using `table._cells`, and the performance improvement is pretty big.

Before adding cache: 99.94s
Added cache: 1.55s

Hopefully this feature will be merged to the master branch because accessing large tables are just too slow here...

Thanks！


Treedy2020:
Very useful bro, how does the magic happen?

PR #1064, opened 2022-03-07
Add api access to `outlineLvl` element of paragraph properties.

lonetwin:
This PR exposes an api to get/set the value for the
[Associated Outline Level](https://c-rex.net/projects/samples/ooxml/e1/Part4/OOXML_P4_DOCX_outlineLvl_topic_ID0EKQCM.html) element.

This element does not directly influence the appearance of text in the
word-processing application and instead is intended to inform the appearance
and `level` of the associated text places like the TOC and the 'Navigation
Pane' (eg: in MS Word the thing that opens up when you click on the page
numbers at the lower left corner of the window).

Co-authored-by: @noddythedon


lonetwin:
Hello @scanny, gentle request -- would you mind reviewing this patchset ? This one's an attempt to get the process right before I submit more, potentially larger, PRs,

Issue #659, opened 2019-04-09
There is any way to insert SVG picture into word document ?


Mkamal88:
I'm trying to insert svg into word document but I got this error:
docx.image.exceptions.UnrecognizedImageError

it seems the function add_picture doesn't support svg format, but it can be done using word UI

madphysicist:
Seems like duplicate of #651.

scanny:
I agree. Closing as duplicate of #651. Thanks @madphysicist :)

goomesthiago:
Is there any update on this? Does python-docx already supports adding svg images into a word document?

Issue #651, opened 2019-04-03
Add svg picture 


kndidox:
Hi guys! Thanks for your great work! It seems that word has been providing support for svg for a couple of years now. I have tried adding an svg file using add_image but I get the following error message: 

> docx.image.exceptions.UnrecognizedImageError.

 I was wondering if you were planning to add support to svgs in the near future. I also tried adding the svg by manipulating the xmls, but I could not make it work either. It was a little too complicated for me. Thanks again!!

scanny:
I don't believe Word supports SVG _natively_. What I believe it does is translate the SVG to EMF (Windows Meta File/Extended Meta File) format and embed that. Since EMF is a vector format, it looks just like the SVG.

The problem is that there is no SVG to EMF translator available in Python, at least the last time I looked. So this functionality (which would also be desirable for `python-pptx`) is not feasible at the moment.

kndidox:
Hi @scanny thanks for your quick answer.

In working in MS Word 2016, as you surely know you can add a svg picture. I have unzipped the Word document to see how they manipulate this kind of files, then i realize that the xml drawing node is very similar for a svg/png. and in the media folder the svg is saved as it without changes, and all the references in the relationship file are to the svg.

So (and this is my supposition) it is technically possible give support to this without convertions.

If i can help you in anything let me know.

BillChen2000:
I'm using Office 365 and I'm sure that Word supports svg format. Encountering the same issue, hope it can be fixed soon.

ftomassetti:
Is this a duplicate of https://github.com/python-openxml/python-docx/issues/505 ?

ftomassetti:
And of https://github.com/python-openxml/python-docx/issues/351

mrgou:
It would be really awesome if this could be done. Word (and other Office applications) now support SVG _natively_ (they're now imported as editable vector images).

juliuscrn:
Would love to see this feature as well.

Issue #351, opened 2017-01-21
Add SVG


mrlarssen:
Is it possible to add SVG to the document? For instance an svg-string such as 

< svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" version="1.1">
   < rect width="200" height="100" stroke="black" stroke-width="6" fill="green" />
</ svg>

would be perfect. Or just a svg file?

scanny:
Can you do that from the Word application UI?

mrlarssen:
I haven't tried it myself, but it seems like it:

https://support.office.com/en-us/article/Insert-SVG-images-in-Microsoft-Office-69f29d39-194a-4072-8c35-dbe5e7ea528c

scanny:
I suspect it translates them into Windows Metafile Format (WMF) and stores them that way. You'd need to try it and then inspect the .docx file (it's a zip archive) and see what the image was that got included. I expect the file extension of the image file name would reliably indicate the underlying file type its stored as.

mrlarssen:
I guess the documentation is outdated, as this approach does not let me insert an svg into the word document. So I assume the limitation lays with word itself?

Then I need to look into how I can convert svg to png in Python.

Thanks for your response.

scanny:
This page has some insights on why it might not be working for you:
https://www.quora.com/How-would-you-insert-a-svg-file-in-PowerPoint

It could be your PowerPoint version is not the very latest; I know mine isn't :)

It's still not clear whether the new PowerPoint 2016 supports SVG files natively or whether they are just converted to WMF/EMF and stored that way, like it does with PDF/EPS images.

ftomassetti:
A PR related to this is pending https://github.com/python-openxml/python-docx/pull/798

kuri65536:
Thank you to python-docx, this is very useful software.

I had done to embed svg images with python-docx.

this is a part of my
[mkdocs-theme-topdf](https://github.com/kuri65536/mkdocs-theme-topdf) .  
[code is here](https://github.com/kuri65536/mkdocs-theme-topdf/blob/main/topdf/docx_svg_hack.py)

BTW, I found the PR and this is just a sample workaround for
the current version.

![Screenshot_20200811-153340~2](https://user-images.githubusercontent.com/11357613/89865323-792b5500-dbe8-11ea-9223-537a32cf602a.png)

(this preview from Word android)

```
from docx.image import image

def run():
    monkey()
    doc = add_paragraph
[...]

PR #196, opened 2015-08-08
EMF image support

eruffaldi:
This patch allows to add EMF images. EMF contains size in 0.01 mm units, so I have picked 300 DPI for computing pixel size. 

Provided also a sample script with EMF for testing

PR #1107, opened 2022-06-24
Add basic support for adding SVG pictures to docx files

takis:
See issues #351, #651, #659.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,12 @@
 Release History
 ---------------
 
+1.1.0 (2023-11-03)
+++++++++++++++++++
+
+- Add BlockItemContainer.iter_inner_content()
+
+
 1.0.1 (2023-10-12)
 ++++++++++++++++++
 
--- a/src/docx/__init__.py
+++ b/src/docx/__init__.py
@@ -13,7 +13,7 @@ from docx.api import Document
 if TYPE_CHECKING:
     from docx.opc.part import Part
 
-__version__ = "1.0.1"
+__version__ = "1.1.0"
 
 
 __all__ = ["Document"]


liam-veitch:
Hi @takis thanks for the improvement - for my SVG files produced by matplotlib there are relative scalings in the width and height attributes. I had to take the data from the viewbox attribute:

Scaling is not great in document, but maybe you know the cause?

Example svg header:
`<svg baseProfile='full' height='100.000000%' version='1.1' viewBox='0.000000 0.000000 1240.000000 1753.000000' width='100.000000%' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'>`

in svg.py:
```
    @classmethod
        stream.seek(0)
        data = stream.read()
        root = ET.fromstring(data)
        # FIXME: The width could be expressed as '4cm'
        # See https://www.w3.org/TR/SVG11/struct.html#NewDocument
        width_str = root.attrib["width"]
        height_str = root.attrib["height"]
        if r'%' in width_str and r'%' in height_str:
            vb_lims = str(root.attrib["viewBox"]).split(' ')
            width_scale = float(width_str.strip(r'%')) * 0.01
            height_scale = float(height_str.strip(r'%')) * 0.01
            width = int(width_scale * float(vb_lims[2]))
            height = int(height_scale * float(vb_lims[3]))
        else:
            width = int(root.attrib["width"])
            height = int(root.attrib["height"])

        return width, height
```

JaverHumberto:
Hi @takis, I'm new using docx (and python), and I'm tryng to load some SVG images in a word document, looking for that I found this, but try to replicate the code you were testing with and it doesn't work for me, I get the docx.image.exceptions.UnrecognizedImageError. So it isn't "included" in docx now? if not, is there a way for using your code? thanks in advance, and sorry if it's not a place for asking this, i'll remove the comment if needed.

Kladdy:
This would be amazing if it could be merged into a new release. In the meantime, if somebody wants to play around with SVG files, I created a monkey patch just as was made in the case for EMF files: https://github.com/python-openxml/python-docx/pull/196#issuecomment-758974499. Just add the code from the gist below to a file called `docx_svg_patch.py`, and at the top of your own python script you have `import docx_svg_patch`. I haven't tested this extensively, and I am not sure how well this plays with other image types now, but it's at least a start.

Link to gist: https://gist.github.com/Kladdy/d3bdb9bbf2c38d4f194ea9a7904fc3f2

takis:
> Hi @takis, I'm new using docx (and python), and I'm tryng to load some SVG images in a word document, looking for that I found this, but try to replicate the code you were testing with and it doesn't work for me, I get the docx.image.exceptions.UnrecognizedImageError. So it isn't "included" in docx now? if not, is there a way for using your code? thanks in advance, and sorry if it's not a place for asking this, i'll remove the comment if needed.

Hi @JaverHumberto, no, unfortunately, the code was not merged. I've created a new pull request and updated the code though. If you want to try it, you can use my fork for now:
https://github.com/takis/python-docx/

takis:
Hi @Kladdy 

I've updated the pull request and I'm also still hoping that it will get merged one day.

takis:
Hi @goomesthiago 

Unfortunately not, but I do try to keep my fork in sync with the main repository.

PR #1118, opened 2022-07-17
Improving Paragraph.text performance by 500%~

yoniy-talon:
During my tests with large docx files with @DrorRoth, we managed to increase Paragraph.text performance by 500%~ using this simple fix.


creyD (approved):
Looks good

cenedella (approved):
Would be great to have these changes accepted into the library.

creyD:
@scanny Could we please get a timeline, when this would be able to be merged or what needs to be adjusted?

lightmare:
You don't even need to create a list, just pass the generator itself to join.

PR #1126, opened 2022-08-06
Fixed a typo for WD_PARAGRAPH_ALIGNMENT

bdgalloway:
Fixed a typo for WD_PARAGRAPH_ALIGNMENT

### DIFF
--- a/docs/user/text.rst
+++ b/docs/user/text.rst
@@ -60,14 +60,14 @@ set to left, centered, right, or fully justified (aligned on both the left
 and right sides) using values from the enumeration
 :ref:`WdParagraphAlignment`::
 
-    >>> from docx.enum.text import WD_ALIGN_PARAGRAPH
+    >>> from docx.enum.text import WD_PARAGRAPH_ALIGNMENT
     >>> document = Document()
     >>> paragraph = document.add_paragraph()
     >>> paragraph_format = paragraph.paragraph_format
 
     >>> paragraph_format.alignment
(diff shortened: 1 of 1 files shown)


DasWordNerd:
@bdgalloway I apologize if I'm completely wrong here--I'm new at participating in these OSS pull request conversations.

My understanding of the `WD_PARAGRAPH_ALIGNMENT` enumeration is that it has an alias of `WD_ALIGN_PARAGRAPH`. My understanding stems from the documentation [here](https://python-docx.readthedocs.io/en/latest/api/enum/WdAlignParagraph.html), and the source for docx > enumerations > text [here](https://github.com/python-openxml/python-docx/blob/36cac78de080d412e9e50d56c2784e33655cad59/docx/enum/text.py#L12).

If the `WD_PARAGRAPH_ALIGNMENT` class's functionality is identical to the functionality of the class's alias `WD_ALIGN_PARAGRAPH`, then the documentation cited in this pull request might not contain a typographical error. Instead, the documentation cited may reflect a style choice by the documentation's author.

Do you agree with the analysis above?

garnik-arut:
I agree with @DasWordNerd.
I assume the author tries to match Word documentation.
[https://learn.microsoft.com/en-gb/office/vba/api/word.wdparagraphalignment](https://learn.microsoft.com/en-gb/office/vba/api/word.wdparagraphalignment)
Part of the problem is that it calls enumeration `WdParagraphAlignment`, but enums themselves have names starting with `wdAlignParagraph`

PR #1259, opened 2023-10-13
Include requirements files in PyPI sdist

musicinmybrain:
These are needed to run tests via `tox.ini`.


scanny:
Included in Release 1.1.0 yesterday.

PR #1270, opened 2023-10-18
settings: get and set on trackRevisions

bpatjuddoc:
**Feature:** Getter and setter for [trackRevisions](https://support.microsoft.com/en-us/office/track-changes-in-word-197ba630-0f5f-4a8e-9a77-3712475e806a) field in settings.xml.

I have a requirement to enable and disable trackRevisions flag on Docx files.
Here is an attempt to do that by enabling getter and setter on [trackRevisions](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.trackrevisions?view=openxml-2.8.1), which is similar to an already existing functionality [evenAndOddHeaders](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.evenandoddheaders?view=openxml-2.8.1). Used the [commit](https://github.com/python-openxml/python-docx/commit/a46dd645f94ad8b3c53878d3dae1a0510661317b) as reference.

### DIFF
--- a/src/docx/oxml/__init__.py
+++ b/src/docx/oxml/__init__.py
@@ -76,6 +76,10 @@ register_element_cls("w:r", CT_R)
 register_element_cls("w:t", CT_Text)
 
 # ---------------------------------------------------------------------------
+# track revisions configurations
+register_element_cls("w:trackRevisions", CT_OnOff)
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


bpatjuddoc:
@scanny - Would really appreciate your thoughts on this - please take a look

Issue #1308, opened 2023-11-29
Enable copying of a run's font


luissantosHCIT:
Hello,

I am new to this project and have been enjoying the work done.
For my own project, I replace some text fields with specific values.
Unfortunately, I noticed that the formatting gets lost when using the Paragraph::text property.
For example,
![image](https://github.com/python-openxml/python-docx/assets/127312735/564a1f02-edb9-4231-96f0-50d2d47cf8b8)
becomes
![image](https://github.com/python-openxml/python-docx/assets/127312735/0a08c70d-3d63-4834-924d-6ff1d7ad3797)

Inspection of the XML output revealed that the run's embedded font information gets lost.
![image](https://github.com/python-openxml/python-docx/assets/127312735/2614999c-e446-4fc0-bf2c-30a9f6677b53)

Finally, I found this comment in the source code for the Paragraph class
![image](https://github.com/python-openxml/python-docx/assets/127312735/b9c1dd73-0c20-4691-afc2-6fcfb7e860e6)

Looking deeper, I see why it the call to clear() was needed.
In my testing I can see how Word can sometimes split text over multiple runs.

As a result, I resorted to logic like so in my program in order to recover the font information.
       ```
if len(p.runs):
            old_run = p.runs[0]
            p.text = p.text.replace(str(key), str(value))
            new_run = p.runs[0]
            new_run.italic = old_run.italic
            new_run.bold = old_run.bold
            new_run.underline = old_run.underline
            new_run.font = old_run.font
            new_run.style = old_run.style
        else:
            p.text = p.text.replace(str(key), str(value))
```

The one issue is that new_run.font = old_run.font was returning an error.

I went ahead and added the missing font.setter and color.setter to the respective properties in run.py and font.py.
I am going to open a pull request to associate with this issue.

Ideally, I can retain the formatting of one of the runs when changing the text via an instance of Paragraph. Since I understand this is tricky and probably not worth the time, I think enabling the ability to copy the font formatting from a run to another would be helpful for those who are assuming 1 formatting to one block of text (meaning, even if Word splits the text into multiple runs, all runs have the same formatting). I think this is a good compromise for me at this time.

Let me know what you think, how I can be of assistance, or how I can adjust my solution to better match expectations in this project.

Thank you for your excellent work!

luissantosHCIT:
I should mention that I tested my changes in #1309 to be working. It allowed me to recover the font information and pass it to the new run generated via Paragraph::text .

[...]

PR #1309, opened 2023-11-29
Enable copying of a run's font

luissantosHCIT:
See issue #1308 for details.

My commits here present my suggested solution.
Feel free to request improvements to solution.

Thank you!

### DIFF
--- a/src/docx/text/font.py
+++ b/src/docx/text/font.py
@@ -53,6 +53,12 @@ class Font(ElementProxy):
         font."""
         return ColorFormat(self._element)
 
+    @color.setter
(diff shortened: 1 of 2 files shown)


luissantosHCIT:
Hello, it has been a long time and no one has commented as to whether this PR has merit.
Since the time I opened, I have noticed that other classes could benefit from having a simple interface to do the formatting copying (all of these rPr kind of objects). 

Should I pursue this further or should I simply have this as a patch set in my systems that I would need to apply post upgrades?
What could I do to get this kind of change to an acceptable state if you think there is benefits to my proposed changes?

Thank you for your time and effort!

PR #1361, opened 2024-03-22
chore: update sphinx

sierzput:


scanny:
Newer versions of Sphinx change the look of the documentation is ways I don't prefer. This is the main reason it hasn't been updated recently.

PR #1378, opened 2024-04-21
Add support for macro enabled template

sunnyar:


sunnyar:
@scanny Is it possible to add this change to support macro enabled .docm word doc files ?

PR #1343, opened 2024-02-22
Add basic support for adding SVG pictures to docx files

takis:
See issues #351, #651, #659.

This replaces this pull request from two years ago:
https://github.com/python-openxml/python-docx/pull/1107

PR #1386, opened 2024-05-02
 Add basic support for adding SVG pictures to docx files #1343 

takis:
See issues https://github.com/python-openxml/python-docx/issues/351, https://github.com/python-openxml/python-docx/issues/651, https://github.com/python-openxml/python-docx/issues/659.

This replaces both earlier pull requests: 
- https://github.com/python-openxml/python-docx/pull/1343: 2024-02-22
- https://github.com/python-openxml/python-docx/pull/1107: two years ago

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,12 +3,20 @@
 Release History
 ---------------
 
+1.2.0 (2025-06-16)
+++++++++++++++++++
+
+- Add support for comments
+- Drop support for Python 3.8, add testing for Python 3.13
+
+
 1.1.2 (2024-05-01)
 ++++++++++++++++++
 
 - Fix #1383 Revert lxml<=4.9.2 pin that breaks Python 3.12 install
 - Fix #1385 Support use of Part._rels by python-docx-template
 - Add support and testing for Python 3.12
+
 
 1.1.1 (2024-04-29)
 ++++++++++++++++++
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 BEHAVE = behave
 MAKE   = make
 PYTHON = python
-BUILD  = $(PYTHON) -m build
 TWINE  = $(PYTHON) -m twine
 
 .PHONY: accept build clean cleandocs coverage docs install opendocs sdist test
@@ -24,10 +23,10 @@ help:
 	@echo "  wheel        generate a binary distribution into dist/"
 
 accept:
-	$(BEHAVE) --stop
+	uv run $(BEHAVE) --stop
 
 build:
-	$(BUILD)
+	uv build
 
 clean:
 	# find . -type f -name \*.pyc -exec rm {} \;
@@ -38,7 +37,7 @@ cleandocs:
 	$(MAKE) -C docs clean
 
 coverage:
-	py.test --cov-report term-missing --cov=docx tests/
+	uv run pytest --cov-report term-missing --cov=docx tests/
 
 docs:
 	$(MAKE) -C docs html
@@ -50,16 +49,16 @@ opendocs:
 	open docs/.build/html/index.html
 
 sdist:
-	$(BUILD) --sdist .
+	uv build --sdist
 
 test:
-	pytest -x
+	uv run pytest -x
 
 test-upload: sdist wheel
-	$(TWINE) upload --repository testpypi dist/*
+	uv run $(TWINE) upload --repository testpypi dist/*
 
 upload: clean sdist wheel
-	$(TWINE) upload dist/*
+	uv run $(TWINE) upload dist/*
 
 wheel:
-	$(BUILD) --wheel .
+	uv build --wheel
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dev = [
     "pytest>=8.4.0",
     "ruff>=0.11.13",
     "tox>=4.26.0",
+    "twine>=6.1.0",
     "types-lxml-multi-subclass>=2025.3.30",
 ]
 
--- a/src/docx/__init__.py
+++ b/src/docx/__init__.py
@@ -13,7 +13,7 @@ from docx.api import Document
 if TYPE_CHECKING:
     from docx.opc.part import Part
 
-__version__ = "1.1.2"
+__version__ = "1.2.0"
 
 
 __all__ = ["Document"]
--- a/uv.lock
+++ b/uv.lock
@@ -21,6 +21,15 @@ wheels = [
 ]
 
 [[package]]
+name = "backports-tarfile"
+version = "1.2.0"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/86/72/cd9b395f25e290e633655a100af28cb253e4393396264a98bd5f5951d50f/backports_tarfile-1.2.0.tar.gz", hash = "sha256:d75e02c268746e1b8144c278978b6e98e85de6ad16f8e4b0844a154557eca991", size = 86406 }
+wheels = [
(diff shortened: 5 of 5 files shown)


marcuoli:
Hello, Is there a target date for this code to be committed to the master ?

I really need this implementation.

Thanks for the awesome software.

takis:
Hi @marcuoli, 

No, unfortunately not. 

I do try to keep my fork in sync with the main repository.

pank:
I hope this patch will be merged soon.

I'm trying to use your patch, but it doesn't appear to work with matplotlib figures since their sizes are [in `pt`](https://github.com/matplotlib/matplotlib/blob/eb812a8e479d0bcf43a30232c31afec78963ed53/lib/matplotlib/backends/backend_svg.py#L320) rather than pixels.

Example: 

``` python
import docx

import numpy as np
import matplotlib.pyplot as plt
from io import BytesIO

from tempfile import mktemp

x = np.linspace(-2*np.pi, 2*np.pi, 100)

fig, ax = plt.subplots()
ax.plot(x, np.sin(x), color="C0")
ax.plot(x, np.cos(x), color="C1")

buf = BytesIO()
fig.savefig(buf, format="svg")
w, h = (Inches(s) for s in fig.get_size_inches())

p = mktemp(suffix=".svg")
fig.savefig(p)

# doesn't work
doc = docx.Document()
doc.add_picture(buf, w, h)
```
Traceback:
```
File c:\Users\rpr\.conda\envs\funandgames\Lib\site-packages\docx\image\svg.py:43, in Svg._dimensions_from_stream(cls, stream)
     40 root = ET.fromstring(data)
     41 # FIXME: The width could be expressed as '4cm'
     42 # See https://www.w3.org/TR/SVG11/struct.html#NewDocument
---> 43 width = int(root.attrib["width"])
     44 height = int(root.attrib["height"])
     45 return width, height
```

I think a pt is 1.33x a pixel in svgs. If so, it might be easy to at least handle pt. But maybe it depends on dpi? 
For matplotlib at least, I think dpi is fixed at 72 (e.g. [here](https://github.com/matplotlib/matplotlib/blob/eb812a8e479d0bcf43a30232c31afec78963ed53/lib/matplotlib/backends/backend_svg.py#L1343))

An example of an mpl svg is attached. 
test

pank:
With this patch I can insert mpl figures. It's written rather verbose, as I don't really know the style of `python-docx`.
@takis would you consider adding this patch or something similar?

```` patch
From 458df2e57c57be410e6d2486ae67e44cebbd177e Mon Sep 17 00:00:00 2001
From: Rasmus <git@pank.eu>
Date: Fri, 18 Oct 2024 10:02:49 +0200
Subject: [PATCH] svg: Handle non-pixels units

---
 src/docx/image/svg.py | 32 ++++++++++++++++++++++++++++----
 1 file changed, 28 insertions(+), 4 deletions(-)

diff --git a/src/docx/image/svg.py b/src/docx/image/svg.py
index 0247a77..4be43e1 100644
--- a/src/docx/image/svg.py
+++ b/src/docx/image/svg.py
@@ -33,13 +33,37 @@ class Svg(BaseImageHeader):
         """
         return "svg"
 
+    @classmethod
+    def _dimension_to_px(cls, s):
+        """Convert SVG unit *s* to pixels.
+
+        Based on the `table`_ on absolute length in W3 CSS Values 3.
+
+        .. table: https://www.w3.org/TR/css-values-3/#absolute-lengths
+        """
+        s = str(s)
+        unit = s.lstrip("0123456789.").casefold()
+        snumber = s[:-len(unit)] if len(unit) > 0 else s
+        number = float(snumber)
+        inch = 96
+        cm = inch/2.54
+        units = {"cm": cm,
+                 "mm": cm/10,
+                 "q": cm/40,  # 'Q' in W3 table
+                 "in": inch,
+                 "pc": inch/6,
+                 "pt": inch/72,
+                 "px": 1}
+        # Unitless is pixels
+        number_px = number * units.get(unit, 1)
+        return int(number_px)
+
     @classmethod
         stream.seek(0)
         data = stream.read()
         root = ET.fromstring(data)
-        # FIXME: The width could be expressed as '4cm'
-        # See https://www.w3.org/TR/SVG11/struct.html#NewDocument
-        width = int(root.attrib["width"])
-        height = int(root.attrib["height"])
+        # The width could be expressed as '4cm'
+        width = cls._dimension_to_px(root.attrib["width"])
+        height = cls._dimension_to_px(root.attrib["height"])
         return width, height
-- 
2.44.0.windows.1
```

lsaint (commented):
This PR broke a unitest which need to add `xmlns:asvg="http://schemas.microsoft.com/office/drawing/2016/SVG/main`
 to
`tests/test_files/snippets/inline.txt`.

some of svg files without "width" and "height" not supported yet.
eg:
`<svg viewBox="0 0 96 96" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" id="Icons_CheckboxCrossed" overflow="hidden"><path d="M37.076 62.948 48.008 52.201 58.939 62.948 63.146 58.67 52.286 47.994 63.146 37.318 58.939 33.04 48.008 43.787 37.076 33.04 32.87 37.318 43.729 47.994 32.87 58.67 37.076 62.948Z"/><path d="M21 21 21 75 75 75 75 21ZM69 69 27 69 27 27 69 27Z"/></svg>`


I have fix them and add some unittests in my [private fork](https://github.com/lsaint/python-docx-oss/pull/27/files#diff-7173581026c8f13f7fe44063d8aa674694785f61ed104f52cd737c231f48db5eR49)

sgtyfianalisisfinanciero:
- Thanks to @pank 's patch, that solved the pt to px issue, now matplotlib svg images can be added, however, it only seems to work sometimes, as some other times I get a placeholder of the image instead. I attach the .svg's at the end if it is helpful.
- The exact same thing occurs when using @lsaint 's fork (that does not solve the pt issue afaik, but supports emf images) with the same images but converted to emf via inkscape
- Copypasting the svg files into the document works well so I think the images itself are not corrupted or something
bolsas
 
divisas

### Original svg images used

preciosos
bolsas_eur
bolsas_usa
divisas
Materias primas

pank:
@sgtyfianalisisfinanciero
Could you share the MPL code of something that doesn't work? Then I will be happy to see if I can reproduce it.

AI've written an package that hotpatches python-docx to support svgs based on this thread that I use for work until svg support is merged. Would that be interesting for others?

sgtyfianalisisfinanciero:
@pank nervermind, the problem was not with the images or the mpl code that generated it. Trying to elaborate a minimal reproducible example I catched the issue

My workflow was something like that:
- Paste images in some table cells (works)
- Generate some other (totally independent) tables, **saving the partial document after each table generation** for debugging purposes
- Paste some other images in some table cells (placeholder shown)

Removing the saving step solves the issue. I'm clueless about why, but it's fine

On the other hand I think your hotfix package would be very interesting for a lot of people. While this get merged, everyone who needed that feature could just `uv add git+...` your fork, which is nicer than hardcoding your patch (which can easily get hard to find in the conversation) into the .venv folder.

alexcraviotto:
How it goes?

ecciv:
This is a really important feature, what are the steps needed prior to a merge? @takis, @lsaint, @scanny

pank:
@sgtyfianalisisfinanciero sorry about the delay. I wanted to add tests and read through the code, but didn't manage before my going on vacation. I still haven't managed, but have added the hotpatch package here:

https://github.com/pank/python-docx-svg

I haven't added it to PyPi, but could do so if that would be useful. It can still be installed via `pip`.

alastaira:
And more news on this? It seems like the patch has addressed the issue, but is still waiting to be merged?

PR #1436, opened 2024-10-01
fix zipfile.py 'UserWarning: Duplicate name: docProps/core.xml'

cip91sk:
Some documents in _rels/.rels, for the relation the targets "docProps/core.xml", have the relationship type
`http://schemas.openxmlformats.org/officedocument/2006/relationships/metadata/core-properties` 
instead of
`http://schemas.openxmlformats.org/package/2006/relationships/metadata/core-properties`

In these cases, python-docx doesn't see that the core-properties are already set, it creates new ones and append a duplicate docProps/core.xml file in the zip

### DIFF
--- a/src/docx/opc/package.py
+++ b/src/docx/opc/package.py
@@ -175,9 +175,12 @@ class OpcPackage:
         try:
             return cast(CorePropertiesPart, self.part_related_by(RT.CORE_PROPERTIES))
         except KeyError:
-            core_properties_part = CorePropertiesPart.default(self)
-            self.relate_to(core_properties_part, RT.CORE_PROPERTIES)
-            return core_properties_part
+            try:
+                return cast(CorePropertiesPart, self.part_related_by(RT.CORE_PROPERTIES_OFFICEDOCUMENT))
+            except KeyError:
+                core_properties_part = CorePropertiesPart.default(self)
+                self.relate_to(core_properties_part, RT.CORE_PROPERTIES)
+                return core_properties_part
 
 
 class Unmarshaller:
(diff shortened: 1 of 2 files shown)


src/docx/opc/package.py
```diff
@@ -175,9 +175,15 @@ def _core_properties_part(self) -> CorePropertiesPart:
         try:
             return cast(CorePropertiesPart, self.part_related_by(RT.CORE_PROPERTIES))
         except KeyError:
-            core_properties_part = CorePropertiesPart.default(self)
-            self.relate_to(core_properties_part, RT.CORE_PROPERTIES)
-            return core_properties_part
+            try:
+                office_document_part = self.part_related_by(RT.CORE_PROPERTIES_OFFICEDOCUMENT)
+                rel = self.relate_to(office_document_part, RT.CORE_PROPERTIES_OFFICEDOCUMENT)
+                self.rels[rel].set_reltype(RT.CORE_PROPERTIES)
```
keestux, line 181: First see below how to create a `setter`. Then change this line to
```python
self.rels[rel].reltype = RT.CORE_PROPERTIES
```

src/docx/opc/rel.py
```diff
@@ -134,6 +134,9 @@ def is_external(self) -> bool:
     def reltype(self) -> str:
         return self._reltype
 
+    def set_reltype(self, reltype:str) -> None:
+        self._reltype = reltype
```
keestux, line 138: This should be a `setter`

```python
    @reltype.setter
    def reltype(self, value: str):
        self._reltype = value
```

rafguns:
Hi @cip91sk , anything I can do to help this PR along? I would add the requested changes myself but I can't (at least not in this PR) since I'm not a maintainer.

cip91sk:
Sorry for the delay, I missed the previous messages.

I corrected with a proper setter, let me know if it's all good now!

Issue #1351, opened 2024-02-29
Error while loading the docx(KeyError: "There is no item named 'word/#_top' in the archive")


akash97715:
Hello Team we are using below code to load the document 

```
from docx import Document

# Path to your DOCX file
docx_file_path = 'myfile.docx'

# Load the DOCX file
document = Document(docx_file_path)

# Example: Print all the text in the document
for para in document.paragraphs:
    print(para.text)
```

Getting below error:

```
[...]

Issue #1105, opened 2022-06-23
Open Word docx file with "The image part with relationship rID8 was not found" error, it always failed


paradise321:
When I try to open a docx file with "The image part with relationship rID8 was not found" error, the following error is reported always.

The code is as simple as following:

```

from docx import Document

doc = Document('.\WordHasSomeProblem.docx')
```
[WordHasSomeProblem.docx](https://github.com/python-openxml/python-docx/files/8966092/WordHasSomeProblem.docx)

The problematic word file is attached,  i.e., WordHasSomeProblem.docx

As Microsoft has declared here: https://docs.microsoft.com/en-us/office/troubleshoot/word/image-part-relationship-rld8-not-found-error-microsoft-word, the problem happens here is because of some target field was set to "NULL.".

My question is: is that possible to add some parameter in Document() to just ignore this kind of errors and continue loading the docx into RAM?

Thanks!

paradise321:
error traces like below:

>  %Run read_word.py
Traceback (most recent call last):
  File "C:\Temp\Source\Python\Word1\read_word.py", line 3, in <module>
    doc = Document('.\WordHasSomeProblem.docx')
  File "C:\Users\userid\AppData\Local\Programs\Python\Python310\lib\site-packages\docx\api.py", line 25, in Document
    document_part = Package.open(docx).main_document_part
  File "C:\Users\userid\AppData\Local\Programs\Python\Python310\lib\site-packages\docx\opc\package.py", line 128, in open
    pkg_reader = PackageReader.from_file(pkg_file)
  File "C:\Users\userid\AppData\Local\Programs\Python\Python310\lib\site-packages\docx\opc\pkgreader.py", line 35, in from_file
    sparts = PackageReader._load_serialized_parts(
  File "C:\Users\userid\AppData\Local\Programs\Python\Python310\lib\site-packages\docx\opc\pkgreader.py", line 69, in _load_serialized_parts
    for partname, blob, reltype, srels in part_walker:
  File "C:\Users\userid\AppData\Local\Programs\Python\Python310\lib\site-packages\docx\opc\pkgreader.py", line 110, in _walk_phys_parts
    for partname, blob, reltype, srels in next_walker:
  File "C:\Users\userid\AppData\Local\Programs\Python\Python310\lib\site-packages\docx\opc\pkgreader.py", line 105, in _walk_phys_parts
    blob = phys_reader.blob_for(partname)
[...]

Issue #902, opened 2020-12-03
Unable to read docx containing pictures linking to internal bookmarks: KeyError: "There is no item named 'word/#MyBookmark' in the archive"


aorsten:
# Bug
The document below contains a picture with a hyperlink to an internal bookmark.

[PictureBookmarks.docx](https://github.com/python-openxml/python-docx/files/5638403/PictureBookmarks.docx)

I get this error message when reading the file using python-docx: `KeyError: "There is no item named 'word/#MyBookmark' in the archive"`

## Stack trace
```
#  Command
self.document = Document(document)

#  Trace (from pytest)
..\..\..\..\..\python-docx\docx\api.py:32: in Document
    document_part = Package.open(docx).main_document_part
..\..\..\..\..\python-docx\docx\opc\package.py:117: in open
    pkg_reader = PackageReader.from_file(pkg_file)
..\..\..\..\..\python-docx\docx\opc\pkgreader.py:37: in from_file
    phys_reader, pkg_srels, content_types
..\..\..\..\..\python-docx\docx\opc\pkgreader.py:74: in _load_serialized_parts
    for partname, blob, reltype, srels in part_walker:
..\..\..\..\..\python-docx\docx\opc\pkgreader.py:119: in _walk_phys_parts
    for partname, blob, reltype, srels in next_walker:
..\..\..\..\..\python-docx\docx\opc\pkgreader.py:114: in _walk_phys_parts
    blob = phys_reader.blob_for(partname)
..\..\..\..\..\python-docx\docx\opc\phys_pkg.py:109: in blob_for
    return self._zipf.read(pack_uri.membername)
C:\...\lib\zipfile.py:1337: in read
    with self.open(name, "r", pwd) as fp:
C:\...\lib\zipfile.py:1375: in open
    zinfo = self.getinfo(name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <zipfile.ZipFile filename='C:\\...\\PictureBookmarks.docx' mode='r'>, name = 'word/#MyBookmark'

    def getinfo(self, name):
        """Return the instance of ZipInfo given 'name'."""
[...]

PR #1498, opened 2025-06-22
Skip null and internal bookmarks links when loading document

michaelromagne:
This PR aims at fixing a recurring issue when loading docx files that have elements pointing to an internal bookmark or a target that is NULL.

Fixed issues with this PR:
- https://github.com/python-openxml/python-docx/issues/902
- https://github.com/python-openxml/python-docx/issues/1105
- https://github.com/python-openxml/python-docx/issues/1351

Docling issue as well (and other projects using docx)
- https://github.com/docling-project/docling/issues/1645

Still, there is also an issue with customXML missing elements [here](https://github.com/docling-project/docling/issues/1645#issuecomment-2938558454) : `KeyError: "There is no item named 'customXML/item5.xml' in the archive"` that is not taken care of, I'll don't know if adding another edge case is a good idea, or if you have a better approach in mind.

### DIFF
--- a/src/docx/opc/pkgreader.py
+++ b/src/docx/opc/pkgreader.py
@@ -250,5 +250,13 @@ class _SerializedRelationships:
         if rels_item_xml is not None:
             rels_elm = parse_xml(rels_item_xml)
             for rel_elm in rels_elm.Relationship_lst:
+                # Null target
+                if rel_elm.target_ref in ("../NULL", "NULL"):
+                    continue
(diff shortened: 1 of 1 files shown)


bartgras:
Hi @scanny 
Do you think someone could approve this PR?
I've tested it on my side and it solved similar issue I had with one of my documents.

SMAntony:
+1
please merge this pr

PR #1514, opened 2025-08-21
Update documents.rst

ibrahim-ogunbiyi:
I think it's best to replace `StringIO` with `BytesIO` since we are working with bytes. This ensures that beginners, or anyone skimming the documentation and copying code, end up with correct and functional code.

### DIFF
--- a/docs/user/documents.rst
+++ b/docs/user/documents.rst
@@ -78,11 +78,11 @@ a document like so::
     # or
 
     with open('foobar.docx', 'rb') as f:
-        source_stream = StringIO(f.read())
+        source_stream = BytesIO(f.read())
     document = Document(source_stream)
     source_stream.close()
     ...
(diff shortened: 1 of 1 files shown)


docs/user/documents.rst
```diff
@@ -78,11 +78,11 @@ a document like so::
     # or
 
     with open('foobar.docx', 'rb') as f:
-        source_stream = StringIO(f.read())
+        source_stream = BytesIO(f.read())
```
scanny, line 81: Hmm, interesting. I'm sure this is an artifact of this documentation being written during the Python 2 era, when `StringIO` actually meant "read bytes".

I'll put it on the shortlist and fix it up next time I'm in there.

ibrahim-ogunbiyi:
Alright noted

PR #1563, opened 2026-06-28
Don't define class-scoped fixtures as instance methods

cjwatson:
This is deprecated as of pytest 9.1; see
https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method.


scanny:
Good catch, thanks for this @cjwatson :)

PR #1596, opened 2026-08-14
Raise PackageNotFoundError for a non-zip stream, not BadZipFile

eeshsaxena:
Opening a document from a *path* that isn't a package raises `PackageNotFoundError`, but opening one from a *stream* that isn't a zip leaks a bare `zipfile.BadZipFile`:

```python
import io, docx
docx.Document(io.BytesIO(b"not a docx"))
# zipfile.BadZipFile: File is not a zip file
```

`PhysPkgReader.__new__` checks `is_zipfile()` for a path and raises `PackageNotFoundError` when it isn't a package, but for a stream it hands the file straight to the zip reader ("assume it's a stream and pass it to Zip reader to sort out"). The zip reader never sorted out the not-a-zip case, so callers catching `PackageNotFoundError` (the documented error for a bad package) miss the stream case.

This catches `BadZipFile` in `_ZipPkgReader` and re-raises `PackageNotFoundError`, so a bad stream and a bad path fail the same way. Valid packages are unaffected (verified a normal open/save/read round-trip still works). Added a test mirroring the existing not-a-package-path test.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise PackageNotFoundError for a non-zip stream, not BadZipFile`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining python-docx!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #1598, opened 2026-08-14
Raise OpcError for a package part with invalid XML instead of a bare lxml error

eeshsaxena:
Opening a `.docx` whose XML is corrupt surfaces a raw `lxml.etree.XMLSyntaxError` out of `Document()` rather than a python-docx error. It happens for the main document part and for the OPC parts:

```python
docx.Document(path_to_docx_with_malformed_document_xml)
# lxml.etree.XMLSyntaxError: Specification mandates value for attribute ...
```

The XML is parsed in two spots while loading a package: `XmlPart.load` (document parts) and the OPC `parse_xml` (`[Content_Types].xml` and `.rels`). I wrapped both so a malformed part raises `OpcError` with the offending partname, which lines up with a non-package file already raising `PackageNotFoundError` (both are `OpcError` subclasses, so `except OpcError` catches either).

Added a unit test for `XmlPart.load` with a malformed blob; it raises XMLSyntaxError on master and passes with the change, and the opc part/oxml tests still pass. Found it by fuzzing `Document()` with corrupted docx packages.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise OpcError for a package part with invalid XML instead of a bare lxml error`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining python-docx!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

Issue #1604, opened 2026-09-07
Paragraph.text setter silently detaches comments (and footnote references) added with 1.2 comment support


Dmitry-Kov:
Since 1.2 added `Document.add_comment()`, assigning `paragraph.text` removes the comment's anchor without any indication. `Paragraph.clear()` drops every child except `w:pPr`, which includes `w:commentRangeStart` / `w:commentRangeEnd` (these are paragraph children, not runs) and the run holding `w:commentReference`. The `<w:comment>` stays in `comments.xml`, anchored to nothing. Word opens the file with no warning and just doesn't show the comment.

Self-contained repro, python-docx 1.2.0:

    from docx import Document
    import zipfile, re

    d = Document()
    p = d.add_paragraph("The fee is EUR 12,000 per milestone.")
    d.add_comment(runs=p.runs[0], text="Confirm this figure before circulation", author="Reviewer")
    d.save("before.docx")

    d = Document("before.docx")
    d.paragraphs[0].text = "The fee is EUR 14,000 per milestone."
    d.save("after.docx")

    for f in ("before.docx", "after.docx"):
        z = zipfile.ZipFile(f)
        doc = z.read("word/document.xml").decode()
        cm = z.read("word/comments.xml").decode()
        print(f, len(re.findall(r"commentRange(Start|End)|commentReference", doc)),
              "markers,", len(re.findall(r"<w:comment ", cm)), "comments")

    # before.docx 3 markers, 1 comments
    # after.docx  0 markers, 1 comments

The same happens to `w:footnoteReference` (the footnote body survives in `footnotes.xml`, the reference in the text is gone) and to any `w:ins`/`w:del` wrapping the runs. Probably related to #1519, which reports comments disappearing on save without a repro — I suspect this is the mechanism.

I understand `.text = ...` is documented as replacing the paragraph's content, and that preserving markers across an arbitrary text replacement is not well-defined. But this is the first thing every code sample and every LLM agent reaches for, and the failure is invisible. Options, from small to large:

1. Keep `w:commentRangeStart`/`w:commentRangeEnd` in `clear()` (they are not content), and re-emit `w:commentReference` for each surviving range at the end of the new run. The comment then anchors to the whole new text, which is what Word does when you retype a commented passage.
2. At minimum, a warning in the `text` setter docstring and in the comments section of the docs that setting text removes comment anchors, footnote references and tracked-change markup in that paragraph.
[...]

PR #1605, opened 2026-09-08
Keep comment anchors when assigning paragraph.text

gyanu2507:
Setting `paragraph.text` runs `clear()`, which dropped `w:commentRangeStart` / `w:commentRangeEnd` and the comment reference run. The comment body stayed in comments.xml with nothing pointing at it.

Collect the previous comment ids and re-mark them on the replacement run. The range then covers that whole run, not the original word span.

This is the comment half of #1604. Footnote references and tracked-change markup still drop on `clear()`.


Dmitry-Kov:
Thanks for picking this up. I tested the logic in this PR against python-docx 1.2.0 and it does fix the comment case: before 3 markers / 1 comment, after `paragraph.text = ...` again 3 markers / 1 comment, with `w:commentRangeStart`, `w:commentRangeEnd` and the reference run correctly wrapped around the new run.

Three notes, in case they're useful before review.

**1. Footnote references and tracked changes are still dropped.** Running the same logic on a document that has a footnote reference in the paragraph:

    footnoteReference: 2 -> 1
    FTN002  footnote id=1 is defined but never referenced - the footnote no longer appears

`w:ins` / `w:del` wrapping the runs go the same way, since `clear()` removes them and nothing re-emits them. So this closes the comment half of #1604 but not the whole issue — might be worth either widening the PR or narrowing the "Fixes" to a note that footnotes and revisions remain open.

**2. The comment range widens to the whole paragraph.** If the comment was anchored to a few words inside the paragraph, it now covers all of the replacement text. I think that's the only sensible behaviour for a full-text replacement, and it matches what Word does when you retype a commented passage, but it's worth stating in the docstring rather than leaving as an implicit side effect — otherwise someone will read "stay anchored" as "anchored to the same span".

**3. Pre-existing bug this PR surfaces:** the reference run gets `rStyle w:val="CommentReference"`, but `add_comment()` never adds that style to `styles.xml`, so the reference mark falls back to the run's default formatting in both Word and LibreOffice with no warning. This comes from `Run.mark_comment_range`, not from your change — it's already there in 1.2.0 whenever you add a comment. The definition Word writes is:

    <w:style w:type="character" w:styleId="CommentReference">
      <w:name w:val="annotation reference"/>
      <w:basedOn w:val="DefaultParagraphFont"/>
      <w:uiPriority w:val="99"/>
      <w:semiHidden/><w:unhideWhenUsed/>
      <w:rPr><w:sz w:val="16"/><w:szCs w:val="16"/></w:rPr>
    </w:style>

Happy to file that as a separate issue if you'd rather keep this PR focused. (Same gap existed in adeu and was fixed in their 3.0.3, for reference.)

Checked with ooxml-integrity against the source document, python-docx 1.2.0, Python 3.9.6, macOS.

gyanu2507:
Thanks for checking this against 1.2.0.

Kept the PR on comments. Footnotes and `w:ins`/`w:del` still go through `clear()`, and I don't want to call #1604 closed. The description now says that.

The docstring now states the range covers the whole replacement run.

The missing CommentReference style is outside this change, as you said. A separate issue is the right place if you want to file it.

